Issue #40, opened 2022-08-04
Create S3 access method for OCM


Skarlso:
Make s3 available as an access method.

```yaml
- access:
      region: eu-central-1
      bucket: skarlso-ocm-bucket
      key: file.tar.gz
      type: S3
    name: s3-server
    relation: external
    type: filesystem
    version: 0.0.1
```

PR #57, opened 2022-08-12
Introduce caching for GitHub access method and add tests for the downloader

Skarlso:
~~This PR will wait until #40 is merged. Then we can rebase on it.~~

This PR does some refactoring around the usage of the `Downloader` and adds blob caching to the Github access method.

Manual Test

```
ocm --cred :type=github --cred :hostname=github.com  --cred token=${GITHUB_TOKEN} download resources ocm-github
github.com/skarlso/ocm-test/0.0.1/private-server: 272 byte(s) written
github.com/skarlso/ocm-test/0.0.1/server: 1594 byte(s) written
```

Without creds:

```
ocm download resources ocm-github-public-only
github.com/skarlso/ocm-test/0.0.1/server: 1594 byte(s) written
```


mandelsoft (dismissed):
This downloader package is no access method, so the place does not seem to be a good choice.
May be we could just put it into the accessio package.

The provided default implementation, is it used somewhere?

yitsushi (approved):

mandelsoft (approved):

Issue #54, opened 2022-08-11
Start using error wrapping with `fmt.Errorf` everywhere


Skarlso:
Instead of:

```go
if err != nil {
    return err
}
```

Do:

```go
if err != nil {
    return fmt.Errorf("add some context where we are atm: %w", err)
}
```

the `%w` will create a wrapped error.

PR #63, opened 2022-08-19
refactor: wrap errors (part 1)

yitsushi:
Part 1 because a we have a lot of unwrapped errors and don't want to create huge PRs.

Under the issue there was a comment about using the `errors.Wrapf` (where
errors` is an ocm package) function instead of `fmt.Errorf`.

I don't even know why that function exists, and that's why I did not use it. The
standard built-in error wrapping with `fmt.Errorf()`, the build-in
`errors.Unwrap()` does exactly the same thing.

Let's see in an example:

```go
package main

import (
  builtinerrors "errors"
  "fmt"

  ocmerrors "github.com/open-component-model/ocm/pkg/errors"
)

type MyError struct{}

func (e MyError) Error() string {
  return "my error"
}

func main() {
  fmt.Println("built-in:")
  builtin()

  fmt.Printf("\n---\n\n")

  fmt.Println("ocm version:")
  ocmversion()
}

func builtin() {
  err1 := MyError{}
  err2 := fmt.Errorf("wrapped: %w", err1)

  fmt.Println(err1)
  fmt.Println(err2)
  fmt.Println(builtinerrors.Unwrap(err2))
}

func ocmversion() {
  err1 := MyError{}
  err2 := ocmerrors.Wrapf(err1, "wrapper")

  fmt.Println(err1)
  fmt.Println(err2)
  fmt.Println(builtinerrors.Unwrap(err2))
}
```

and the output:

```
built-in:
my error
wrapped: my error
my error

---

ocm version:
my error
wrapper: my error
my error
```

The only two reason to use `Wrapf`:
1. it forces the standard `msg: err` format.
2. it handles nil automatically and returns with a nil if the error was nil.

The format shouldn't be an issue, that's standard and that's why we have pull
request reviews.

On the 2nd point.

Let's see what that function does:

```go
type wrappedError struct {
  wrapped error
  msg     string
}

func Wrapf(err error, msg string, args ...interface{}) error {
    if err == nil {
        return nil
    }

    if len(args) > 0 {
        msg = fmt.Sprintf(msg, args...)
    }

    return &wrappedError{
        wrapped: err,
        msg:     msg,
    }
}

func (e *wrappedError) Error() string {
    return fmt.Sprintf("%s: %s", e.msg, e.wrapped)
}

func (e *wrappedError) Unwrap() error {
    return e.wrapped
}
```

That whole cose can be replaced by:

```go
func Wrapf(err error, msg string, args ...interface{}) error {
    if err == nil {
        return nil
    }

    if len(args) > 0 {
        msg = fmt.Sprintf(msg, args...)
    }

    return fmt.Errorf("%s: %w", msg, err)
}
```

We could remove the whole `wrappedError` struct.

So technically we can use `Wrapf`

```go
return Wrapf(err, "unable to read '%s'", filename)
```

or use built-in options:

```go
if err != nil {
  return fmt.Errorf("unable to read '%s': %w", filename, err)
}

return nil
```

in both cases we get the same output:

```
unable to read '/tmp/stuff': file not found
```

We could save 3 lines of code and, but we have a code we have to maintain, and
anyone who will read the `Wrapf` line has to check what it does to fully
understand what's happening.

References:
* Built-in errors package: https://pkg.go.dev/errors#Unwrap
* Issue: https://github.com/gardener/ocm/issues/54

Related to #54

```other developer

```

### DIFF
--- a/pkg/common/accessio/access.go
+++ b/pkg/common/accessio/access.go
@@ -16,6 +16,7 @@ package accessio
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"sync"
 
@@ -379,14 +380,22 @@ func (b *blobAccessView) IsClosed() bool {
 func (b *blobAccessView) Get() (result []byte, err error) {
 	return result, b.view.Execute(func() error {
 		result, err = b.access.Get()
-		return err
+		if err != nil {
+			return fmt.Errorf("unable to get access: %w", err)
+		}
+
+		return nil
 	})
 }
 
 func (b *blobAccessView) Reader() (result io.ReadCloser, err error) {
 	return result, b.view.Execute(func() error {
 		result, err = b.access.Reader()
-		return err
+		if err != nil {
+			return fmt.Errorf("unable to read access: %w", err)
+		}
+
+		return nil
 	})
 }
 
--- a/pkg/common/accessio/cache.go
+++ b/pkg/common/accessio/cache.go
@@ -73,15 +73,23 @@ func (c *refMgmt) Unref() error {
 	if c.closed {
 		return ErrClosed
 	}
-	c.refcount--
+
 	var err error
+
+	c.refcount--
 	if c.refcount <= 0 {
 		if c.cleanup != nil {
 			err = c.cleanup()
 		}
+
 		c.closed = true
 	}
-	return err
+
+	if err != nil {
+		return fmt.Errorf("unable to unref: %w", err)
+	}
+
+	return nil
 }
 
 func (c *refMgmt) UnrefLast() error {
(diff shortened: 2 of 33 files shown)


Skarlso (approved):
👍

mandelsoft (changes requested):

mandelsoft (commented):
In most cases, where you added a newline, they are formally required by those strange newline rules, but they break the codeblock that semantically belong together. I'm definitely not a fan of formal rules that are useless in most of the cases..

most of these *if return wrap else return* boilerplate blow up the code. The could be simplified in one line by the `Wrapf` function. This was one reason, why I added it. This way the discussion whether newline is useful here or not, could completely be avoided.

mandelsoft (approved):

PR #69, opened 2022-08-26
OCM descriptor as CRD

### LINKED ISSUE #35: Use a K8s custom resource definition to describe the component descriptor
**Description**
As discussed in several meetings, we should make use of K8s CRD to model an OCM component descriptor, which would enable us to reuse some of the existing tools in the K8s space dealing with CRDs. This CRD is then the basis for V3 component descriptors (in contrast to V2, the currently used one).

**Done Criteria**
- [ ] CRD representing a complete Component Descriptor available

Skarlso:

Fixes https://github.com/gardener/ocm/issues/35

**Release note**:

```feature user

```


mandelsoft (commented):
Is it possible to put the CRD stuff into a some sub directory?
It is quite prominent, so far. And the structure does not really meet the actual project structure.

Basically, instead of using own types for the various elements the CRD should use the 
already existing binding types. Is this possible?

Or alternatively, we need a transfer function, that provides a component descriptor (according to this library) from the K8S resource, and the opposite direction, putting a CD into an K8S resource.

This should be based on the internal structure (package compdesc), not the serialization versions. This would support a smooth usage together with the rest of the lib.

yitsushi (approved):

nikimanoledaki (approved):

PR #77, opened 2022-08-30
Turn on some linters

### LINKED ISSUE #76: Enable linters with the least amount of code changes

yitsushi:
**Enabled all linters with:**
* no issues
* manageable amount of issues
* auto-fixable issues (`--fix` flag), lot of the changes are auto-fix

Started to address some of the big ones too because the output listed
only 15-20 issues first and that's not bad, but after fixed them it
listed another 15-20, so I just disabled them.

**Most of the disabled linters have comment:**
* why I disabled it
* why we should address those issues (for example security, stability)

**Disabled linters are grouped together based on my initial thoughts:**
* Logical next step: we should consider enabling these
* Opinionated: totally opinionated linters or too strict, we can ignore them.
* Disabled with reason: The reason listed next to them.
* Disabled because of generics in go 1.18: The title is self-explanatory
* Deprecated: Deprecated linters, some just abandoned, and some has got
  replaced with a newer and more maintained linters.

**Configuration:**
Configures a few linters to make them more reasonable.

**Exclude list:**
Added some excluded items, for example:
* never check line length, even if we enable it, if the line contains a URL.
* if govet find a shadowed `err` variable, ignore that.
* If a comment has a URL, ignore godox. The reason is simple, we should
  never leave a comment in the code that has "TODO" or "FIXME" without
  creating a GitHub issue, otherwise it will never be fixed.

**Right now, all the running/enabled linters:**
* asasalint
* asciicheck
* bidichk
* decorder
* depguard
* dogsled
* durationcheck
* errchkjson
* errname
* errorlint
* execinquery
* forbidigo
* gci
* goconst
* godot
* gofmt
* gofumpt
* goheader
* goimports
* gomodguard
* goprintffuncname
* gosimple
* govet
* grouper
* importas
* ineffassign
* makezero
* misspell
* nakedret
* noctx
* nolintlint
* nosprintfhostport
* predeclared
* promlinter
* reassign
* staticcheck
* tenv
* thelper
* tparallel
* typecheck
* unconvert
* unused
* usestdlibvars
* whitespace

Closes #76

```other developer

```


mandelsoft (commented):
As I mentioned, before, ordering and structuring the imports is a good idea.

But I definitely cannot understand why the ginkgo/gomega imports should be mixed up with the module local imports.

At least they should be put into the group of external imports, but even better they should be, as before, a separate group.

It must be possible to use own import groups (which for sure, must internally be sorted then)

Then, I'm not sure what newlines have been enforced by the linter, and which ones were injected by you. It seems to be quite inconsistent, how some were introduced. There are for sure much more locations, where the same rules should apply, but no newline was inserted (Don't misunderstand, it is completely fine, that they were NOT inserted) .

BTW: the proposed deleted newline are fine :-)

Skarlso (dismissed):

mandelsoft (approved):
Please, see my additional comments ts on the unresolved conversations above.

Skarlso (approved):

PR #104, opened 2022-09-08
Introduce logging - Part 1 - Basic, globally configurable logging

Skarlso:

Relates to https://github.com/gardener/ocm/issues/64

**Release note**:

```feature user

```


yitsushi (dismissed):
I don't think that's the right way long term, but for now it's still better as it allows to silence stdout.

Long term we should introduce some localized logging that can be configured. Logrus is a good choice and I love that, but with this approach all logging depends on global configuration and not local. The logical and elegant way would be to add a logger option to the context.

yitsushi (dismissed):

jschicktanz (changes requested):
just some little things :)

jschicktanz (approved):

yitsushi (approved):

Issue #144, opened 2022-10-12
OCM CLI Usability Enhancements


phoban01:
**What would you like to be added**:

If we wish to encourage adoption of the OCI specification and tooling there are some usability issues that we should address. The purpose here is to improve the developer experience using the OCM CLI.

The cumulative effect of the improvements outlined here will streamline the ocm cli workflow and should mean that it is not necessary to use a Makefile in order to execute a simple series of ocm commands. This will encourage adoption and enable straightforward integration into build processes and scripted workflows.

- [Component Initialization Process](component-initialization-process)
- [Target Directory](target-directory)
- [Target Parameter](target-parameter)
- [Target Parameter](target-parameter)
- [Creating Resources, Sources & References](creating-resources-sources-references)
- [Aliases & Inversions](aliases-inversions)
- [Unnamed Parameters](unnamed-parameters)
- [Examples](examples)
- [Surface important functionality](surface-important-functionality)

An example workflow with these changes incorporated could the as follows:

```bash!
# initialise our ocm component
$ ocm init github.com/torvalds/cloud-native-kernel --version v1.0.0

# examine the resulting "workspace" and component archive
$ tree
.
├── .ocm/
│   ├── blobs/
│   └── component-descriptor.yaml
    
# create a resource for the podinfo container image
$ ocm create resource podinfo \
    --type ociImage \
    --version 6.2.0 \
    --access-type ociArtefact \
    --access-image-reference ghcr.io/stefanprodan/podinfo:6.2.0

# sign the component version locally with the supplied key
$ ocm sign --local --key ocm-private.key --name devsig

# transfer the component to the github container registry and also move the podinfo resource
$ ocm transfer component --with-resources ghcr.io/linux

# verify after transfer
$ ocm verify --repository ghcr.io/linux github.com/torvalds/cloud-native-kernel:v1.0.0 

# sign the component version remotely with the supplied key
$ ocm sign --key private.key --name ownersig github.com/torvalds/cloud-native-kernel:v1.0.0
```

## Component Initialization Process

1  OCM initialisation workflow needs to be cleaner. An `ocm init` command should replace the `ocm create componentarchive` and be configurable with a single argument with options provided via named flags. 

The current command is as follows:
```bash
ocm create componentarchive github.com/phoban01/ocm-example v1.0.0 phoban01 out
```

The proposed `ocm init` command could be used as follows:

For example:
```bash
ocm init github.com/phoban01/ocm-example --version v1.0.0
```

[...]

PR #154, opened 2022-10-14
enable adding (re)sources directly via command line options

mandelsoft:

As an alternative to adding resources/sources by specifications in a separate source file now three options can be used
to completely describe a single (re)source via command line options (#144):

`--resource`/`--source`/`--reference`: to describe the meta data
`--input`: to describe the data used to compose a resource stored locally with the component version
`--access`: to describe the access specification for referenced artifacts

The data for the meta data is partly fixed, but potentially very complex, therefore it does not seem to be appropriate to describe these attributes by dedicated options. The data for the input and access highly depends on the structure of the used types. The command just uses a dynamic deserialization as part of the currently used specification file and a validation provided by registered implementations. Therefore it is not really possible to find dedicated options for all the possible variants.

Because of this, all three options accept a json/yaml given as string. Only this allows specifying the required arbitrary and potentially deep data structures.

Additionally, there are dedicated options for the meta data, like `--name`, `--version`, and `--type` for (re) sources and `--component` for references. This allows specifying basic meta data with no labels or extra identities just by command line options. Both variants can be combined: specify required meta data per dedicated value option and the rest optionally via the YAML based option.

Fixes #

To handle the dynamic nature of the configuration of access method and input specifications required for the artifact composition commands a new type of option handling has been introduced. With OptionTypes, OptionTypeSets and ConfigProviders based on OptionTypes is it possible now for dynamically registered elements to compose options sets and to  validate given option sets and  finally map given values to generic configuration objects.

This feature is then used by the dynamic parts to register appropriate handlers used by the CLI commands.
They now are completely based on this mechanism to handle meta data of the various element types (resources, sources, references) to be added to a component version, and the content offered either by inputs or access specifications.

Additionally, commands based on the same mechanism has been introduced to configure appropriate 
element specification source files (resources.yaml, sources.yaml), which can then be used as traditional control files for the (component version) add commands.

**Release note**:
It is possible now to describe artifacts or references to be added to a component version completely or partly
by options of the `add` commands without the use of an element description file.

For example:
```bash
$ ocm add resources  /tmp/ca --name image --version  v0.1.0 --type ociImage --accessType ociArtefact --reference ghcr.io/mandelsoft/pause:v0.1.0

$ ocm add resources  /tmp/ca --name image --version  v0.1.0 --type ociImage --inputType docker --hint mandelsoft/pause:v0.1.0 --inputPath mandelsoft/pause:v0.1.0

$ ocm add references  /tmp/ca --name testdata --component  github.com/mandelsoft/pause  --version 1.0 --label "*purpose=test"  --label 'label@v1={"local": true}' 
```

Additionally there are commands accepting the same options to compose an element description file usable for the
add commands:
```bash
$ ocm add resourceconfig  resources.yaml --name myresource --type PlainText --input '{ "type": "file", "path": "testdata/testcontent", "mediaType": "text/plain" }' 
```

```feature user

```


jensh007 (approved):

mandelsoft:
This PR is corrupted since the repository has been set to public. It does not get new commits from the fork and the fork is completely disconnected. 

I had to create a new fork and restore the content from my local local repo. The replacement PR is #161.

PR #162, opened 2022-10-25
differentiate between docker repository name and hostname

mandelsoft:

The legacy notation for dockerhub based references uses a name restriction for the repository
which can be used to differentiate between dockerhub references and flat image names stored in some
other OCI registry.

Fixes #

**Release note**:

### DIFF
--- a/pkg/contexts/oci/grammar/grammar.go
+++ b/pkg/contexts/oci/grammar/grammar.go
@@ -57,6 +57,18 @@ var (
 	// dashes.
 	separatorRegexp = Match(`(?:[._]|__|[-]*)`)
 
+	// dockerOrgSeparatorRegexp defines the separators allowed to be
+	// embedded in a docker organization name.
+	// https://docs.docker.com/docker-hub/repos/
+	dockerOrgSeparatorRegexp = Match(`(?:_|__|[-]*)`)
+
+	// DockerOrgRegexp restricts registry path component names to start
+	// with at least one letter or number, with following parts able to be
+	// separated by one or two underscore and multiple dashes.
+	DockerOrgRegexp = Sequence(
+		AlphaNumericRegexp,
+		Optional(Repeated(dockerOrgSeparatorRegexp, AlphaNumericRegexp)))
+
 	// NameComponentRegexp restricts registry path component names to start
 	// with at least one letter or number, with following parts able to be
 	// separated by one period, one or two underscore and multiple dashes.
@@ -155,7 +167,7 @@ var (
 
 	// DockerReferenceRegexp is a shortened docker reference.
 	DockerReferenceRegexp = Anchored(
-		Capture(NameComponentRegexp, RepositorySeparatorRegexp, NameComponentRegexp),
+		Capture(DockerOrgRegexp, RepositorySeparatorRegexp, NameComponentRegexp),
 		CapturedVersionRegexp)
 
 	TypedRepoRegexp = Anchored(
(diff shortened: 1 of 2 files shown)


pkg/contexts/oci/grammar/grammar.go
```diff
@@ -55,8 +55,21 @@ var (
 	// separatorRegexp defines the separators allowed to be embedded in name
 	// components. This allow one period, one or two underscore and multiple
 	// dashes.
+
 	separatorRegexp = Match(`(?:[._]|__|[-]*)`)
 
+	// dockerRepositorySeparatorRegexp defines the separators allowed to be
+	// embedded in a docker repository name.
+	// https://docs.docker.com/docker-hub/repos/
+	dockerRepositorySeparatorRegexp = Match(`(?:[_-]*)`)
```
yitsushi, line 64: This allows repeated `_` or `-` characters, is this the intention here? So for example `______----___` is a valid separator.

If yes, wouldn't it be easier to just write a general regexp for these cases, instead of using `Repeated` and `Optional` function calls? Something like `^[a-z0-9][a-z0-9_-]*$`, that would allow everything that starts with a letter or number, and followed by any number of letters, numbers, `_`, or `-` characters.

mandelsoft: I just assumed, that the name should at least start end end with a regular identifier part. Do you have a better reference?

yitsushi: No I don't, just asked if that's intentional or not, to make it clear.

yitsushi (changes requested):
tested, it can't contain (on dockerhub) repeated special characters

image

image

image

pkg/contexts/oci/grammar/grammar.go
yitsushi: Added a blocker on the PR, tested it on `hub.docker.com`, tried to create a new repo and it does not like repeated special characters

mandelsoft: I took the initial parsing from the docker library, here they explicitly allow two underscores for repository components.
I'll adapt this by disallowing the dot for the first component of the reference in case of a docker reference. The it should in sync with the parsing done in the docker lib.

yitsushi: Can you please link the relevant moby source code? Thank you.

mandelsoft: If the org parts contains a `.` it us used has hostname and it is not a dockerhub ref.

yitsushi: I still couldn't find the relevant Moby source code :(

yitsushi: Based on that, valid:
* `a.1.c`
* `a-1-c`
* `a_1_c`
* `a__1__c`
* `a------1----c`

and not valid:
* `a..1..c`
* `a._1._c`
* `a-_1-_c`
* `a_-.1_-.c`
* `a___1___c`

yitsushi: Relevant lines:
```
name                            := [domain '/'] path-component ['/' path-component]*
path-component                  := alpha-numeric [separator alpha-numeric]*
alpha-numeric                   := /[a-z0-9]+/
separator                       := /[_.]|__|[-]*/
```

mandelsoft: yes, that is what I've changed now to discover dockerhub references.

mandelsoft: I added some tests for this

yitsushi (approved):

pkg/contexts/oci/grammar/grammar.go
yitsushi: I think not if you put the hostname in from of the org. For example `docker.io/some.org/fancy.image-name:digest`, docker hub does not support special characters in the org part, but does in the repo name, which is still the `path-component`. I assume this regexp is called on the whole name `org/repo`, and org does not allow `.`, the repo name can still contain `.` and with other registries (self hosted) the url will look like this:
```
registry-server.internal/fancy.org/with.fancy-repo
```

and that should be valid  based on the Moby implementation (which is docker itself).
```

PR #163, opened 2022-10-26
generalize a common handling for optional arguments

mandelsoft:

There are several implementation to handle a variadic parameters to default an optional argument.
Using generics these different implementations can be harmonized by central oe.
This assures always the same behaviour of the currently different implementations.

Fixes #

Unfortunately with go generics it is not possible to differentiate between value types and nil-able
types. 

If we choose `comparable` as type parameters which would include pointer types, `interfaces` cannot be passed.
If we choose `any` we cannot compare and would need `reflect.DeepEqual` to select the first non-zero element.

Therefore this generalization is not possible without using reflection and various variants. which makes it a little bit ugly.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/mandelsoft/vfs/pkg/vfs"
+	"github.com/modern-go/reflect2"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/yaml"
 
@@ -254,4 +255,46 @@ func StringMapKeys(m interface{}) []string {
 	}
 	sort.Strings(keys)
 	return keys
+}
+
+// Optional returns the first optional non-zero element given as variadic argument,
+// if given, or the zero element as default.
+func Optional[T any](list ...T) T {
+	var zero T
+	for _, e := range list {
+		if !reflect.DeepEqual(e, zero) {
+			return e
+		}
+	}
+	return zero
+}
+
+// OptionalDefaulted returns the first optional non-nil element given as variadic
+// argument, or the given default element. For value types a given zero
+// argument is excepted, also.
+func OptionalDefaulted[T any](def T, list ...T) T {
+	for _, e := range list {
+		if !reflect2.IsNil(e) {
+			return e
+		}
+	}
+	return def
+}
+
+// OptionalDefaultedBool checks all args for true. If no true is given
+// the given default is returned.
+func OptionalDefaultedBool(def bool, list ...bool) bool {
+	for _, e := range list {
+		if e {
+			return e
+		}
+	}
+	return def
+}
+
+// GetOptionFlag returns the flag value used to set a bool option
+// based on optionally specified explicit value(s).
+// The default value is to enable the option (true).
+func GetOptionFlag(list ...bool) bool {
+	return OptionalDefaultedBool(len(list) == 0, list...)
 }
--- a/cmds/ocm/commands/common/options/closureoption/option.go
+++ b/cmds/ocm/commands/common/options/closureoption/option.go
@@ -15,6 +15,7 @@ import (
 	"github.com/open-component-model/ocm/pkg/common"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/transfer/transferhandler"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/transfer/transferhandler/standard"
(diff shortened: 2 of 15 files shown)


docs/reference/ocm_add_sources.md
```diff
@@ -483,7 +483,7 @@ are handled:
     The size of the blob
   
   
-  Options used to configure fields: <code>--mediaType</code>, <code>--size</code>, <code>--digest</code>, <code>--reference</code>
+  Options used to configure fields: <code>--size</code>, <code>--digest</code>, <code>--reference</code>, <code>--mediaType</code>
```
yitsushi, line 486: Can we fix the generated stuff to be consistent, or run only when we have relevant changes? (just open a new issue about it)

pkg/utils/utils.go
```diff
@@ -255,3 +256,45 @@ func StringMapKeys(m interface{}) []string {
...
+
+// Optional returns the first optional non-zero element given as variadic argument,
+// if given, or the zero element as default.
+func Optional[T any](list ...T) T {
+	var zero T
+	for _, e := range list {
+		if !reflect.DeepEqual(e, zero) {
+			return e
+		}
+	}
```
yitsushi, line 268: That changes the original logic, is it intentional?

* Original logic: Return the first entry if we have at least one
* New logic: Return with the first non-zero entry

We get different result from:
* `[nil, nil, nil, *obj]`
  * old return value: `nil`
  * new return value: `*obj`
* `[Data{}, Data{Name: "name"}, Data{Name: "name2"}]`
  * old return value: `Data{}`
  * new return value: `Data{Name: "name"}`
* `[false, true, true, false]`
  * old return value: `false`
  * new return value: `true`

pkg/utils/utils.go
```diff
@@ -255,3 +256,45 @@ func StringMapKeys(m interface{}) []string {
...
+// OptionalDefaultedBool checks all args for true. If no true is given
+// the given default is returned.
+func OptionalDefaultedBool(def bool, list ...bool) bool {
+	for _, e := range list {
+		if e {
+			return e
+		}
+	}
+	return def
+}
```
yitsushi, line 293: I think based on the new implementation it's equivalent to `Optional(...)` if the argument list is a boolean list, because `var zero T` will be `false` and because of `!reflect.DeepEqual(e, zero)`, it will return `true` only if at least one entry is `true`.

yitsushi (changes requested):
Question about a logical change

docs/reference/ocm_add_sources.md
mandelsoft: That's a good point, it is sign for an error in the old code. It seems to iterate over a map, which doesn't not provide a stable order, so every generation causes another order. I'll fix this.

pkg/utils/utils.go
mandelsoft: yes, that was a problem with these various implementations, every one worked slightly different. Now I've aligned it, and it works the same way everywhere.

pkg/utils/utils.go
mandelsoft: yes, but the default value is different.  And it is not the same behaviour like `OptionalDefaulted` because bool is not nil-able, and if should provide true, if one argument is true.

yitsushi: Well, I don't really see the reason to use `OptionalDefaultedBool(true, ...)` at all. That just means `true` in all cases. So the only use case is `OptionalDefaultedBool(false, ...)` and in that case it's the same as `Optional(...)` with a bool list.

Why I don't see practical?

We have two cases:

1. `OptionalDefaultedBool(true, list...)` where `list` contains a `true` value. In this case the function will return `true`
2. `OptionalDefaultedBool(false, list...)` where `list` contains only `false` values. In this case, the function will return `true`.

So in any cases, if the first argument is `true`, it will always return `true`.

pkg/utils/utils.go
yitsushi: The only real difference is how booleans are handled. Previously if the first entry in a list of booleans was a false, the return value was false.

pkg/utils/utils.go
mandelsoft: Yes, that is correct, but if the default is determined by an expression and not set by a constant, it makes sense again.
Just have a look at the method following this one (`GetOptionFlag`).

The cases are different:
- for 1) it returns always true, regardless of the given additional arguments
- for 2) it returns true at least one true is given, otherwise it returns false.

yitsushi (approved):

PR #165, opened 2022-10-26
rename package core to internal

mandelsoft:

Core packages are intended for internal use, only. There is a concept in GO that prevents the usage of packages named `internal` by non-sibling packages. This is exactly the way the core packages are used.

Therefore the core packages should be renamed accordingly to enforce the usage pattern by the GO language.

 
Fixes #

**Release note**:

```feature user

```


yitsushi (commented):
Can't we just move "internal" code under a root level `internal` directory instead of pkg? The new path would be:

```diff
- pkg/contexts/clictx/core/builder.go
+ internal/contexts/clictx/core/builder.go
```

It feels more natural and easier to see what can be used and what not.

That's why a lot of project has a root level internal directory for example:
- https://github.com/golang/tools/tree/master/internal
- https://github.com/helm/helm/tree/main/internal
- https://github.com/hashicorp/terraform/tree/main/internal
- https://github.com/weaveworks-liquidmetal/flintlock/tree/main/internal
- https://github.com/weaveworks-liquidmetal/cluster-api-provider-microvm/tree/main/internal

yitsushi (approved):

PR #176, opened 2022-11-07
refactor: remove CRD definition for the Component Descriptor

### LINKED ISSUE #171: Clear out the ComponentDescriptor api definition from this repository

The CRD definition for the Component Descriptor is being transferred to the controllers. So it can be completely deleted from this repository.

**Why is this needed**:

yitsushi:
CRDs are defined in the relevant controller repo.

References:
- https://github.com/open-component-model/ocm-controller/tree/main/api/v1alpha1
- https://github.com/open-component-model/ocm/issues/171

Closes #171


mandelsoft (commented):
Do I understand it correctly, the new CRD types in the controller repo do not use the metav1 types from the ocm library component descriptor package anymore?

So, the OCM CD functions cannot be used with the CD part of the CRD spec anymore?

Skarlso:
> Do I understand it correctly, the new CRD types in the controller repo do not use the metav1 types from the ocm library component descriptor package anymore?

They do use it. The controllers import the OCM library. It's the same spec but lives in a different repo where the usage is clearer.

mandelsoft:
Thanks, I was just asking, because the kubebilder annotations have been removed from the metav1 structs. I thought, they were required for the usage of those types in CRD definitions.

mandelsoft (approved):

Skarlso:
> Thanks, I was just asking, because the kubebilder annotations have been removed from the metav1 structs. I thought, they were required for the usage of those types in CRD definitions.

That's a good point. It does work, I also thought it was required. :) I'll double check.

PR #177, opened 2022-11-08
fix: enable linter: maintidx

### LINKED ISSUE #91: enable linter: maintidx
> maintidx measures the maintainability index of each function.

Default is 20, which is too low I think, but we have some extreme functions, for example:
```
name: push, Cyclomatic Complexity: 35
name: apply, Cyclomatic Complexity: 51
```

yitsushi:
Issues fixed:
- Fix issues with auto fix (`--fix`).
- Rename `this` receivers.
- Fix language issues, most of the duplicated words like "the the".
- Removed unnecessary type conversion.
- Reduced complexity of the `_apply` function by extracting some parts into a separate function.
- Reduced complexity of the `push` function by extracting some parts into a separate function.

References:
- https://github.com/open-component-model/ocm/issues/91

Closes #91


yitsushi:
Please review the extracted functions carefully, I tried to be careful, but it does not mean I did not miss something. These functions have no tests, so I couldn't verify everything with tests.
```
❯ go test ./pkg/docker/...
?       github.com/open-component-model/ocm/pkg/docker  [no test files]
?       github.com/open-component-model/ocm/pkg/docker/resolve  [no test files]
```

pkg/docker/pusher.go
```diff
@@ -140,102 +140,15 @@ func (p dockerPusher) push(ctx context.Context, desc ocispec.Descriptor, src res
 		req = p.request(host, http.MethodPut, putPath...)
 		req.header.Add("Content-Type", desc.MediaType)
 	} else {
-		// Start upload request
```
mandelsoft, line 143: I completely understand, why you restructured this code. The change looks ok.
But this code has been copied from  `github.com/containerd/containerd remotes/docker` (see README) to fix the retry behavior.
I would like to keep it as close as possible to the original to make it easier to handle changes in the original version.

cmds/ocm/commands/ocmcmds/common/resources.go
```diff
@@ -443,12 +446,12 @@ func (o *ResourceAdderCommand) ProcessResourceDescriptions(listkey string, h Res
 	defer obj.Close()
 
 	for _, r := range resources {
-		ictx := ictx.Section("adding %s...", r.Spec().Info())
+		isctx := ictx.Section("adding %s...", r.Spec().Info())
```
mandelsoft, line 449: The idea here was, to always have an input context (`ictx`) available, using the same name, regardless of the nesting level.

Therefore at the beginning of a nesting level a new local variable with the same name is created.
This makes the code look the same regardless of the level.

Is this renaming really required by the new linter?

cmds/ocm/topics/toi/bootstrapping/topic.go
```diff
@@ -332,7 +332,7 @@ execution and reading provided executor outputs after the execution.
 /
 └── toi
     ├── inputs
-    │   ├── config      config info from package specification
+    │   ├── config      info from package specification
```
mandelsoft, line 335: I would omit `info` instead of `config`: the file `config` contains the configuration from the package specification.

May be, replace it by `configuration from package specification`to avoid the repetition of `config`.

pkg/contexts/oci/ociutils/helm/ignore/doc.go
```diff
@@ -28,7 +28,7 @@ The formatting rules are as follows:
 
 Example:
 
-	# Match any file named foo.txt
+	# Match exact file name
```
mandelsoft, line 31: This is generated text by helm.

It means, that any occurrence of a file named `foo.txt` is matched, in contrast to /foo.txt, which would match the exact file. I think, this is what this comment want to describe.

pkg/contexts/ocm/accessmethods/localfsblob/method.go
```diff
@@ -11,8 +11,10 @@ import (
 )
 
 // Type is the access type of a blob in a local filesystem.
-const Type = "localFilesystemBlob"
-const TypeV1 = Type + "/v1"
+const (
+	Type   = "localFilesystemBlob"
+	TypeV1 = Type + "/v1"
```
mandelsoft, line 16: If we change it anyway, could you use `TypeV1 = Type + runtime.VersionSeparator + "v1"` similar to the other occurrences.

mandelsoft (changes requested):
Thanks, for all those typo fixes. This was for sure no fun. Just some minor things...

cmds/ocm/commands/ocmcmds/common/resources.go
yitsushi: Yes, and it can be misleading as well, not just for linters, but in general. Shadowing a variable can be tricky and annoying, because as a reader (or someone who wants to change something), they would expect it to be exactly the same, but it's not, it's a new input context with a new section (based on the previous input context).

cmds/ocm/topics/toi/bootstrapping/topic.go
yitsushi:
```suggestion
    │   ├── config      configuration from package specification
```

pkg/contexts/oci/ociutils/helm/ignore/doc.go
yitsushi: I would argue with that, if we go down the meaning of the words, `/foo.txt` is an exact path, while `foo.txt` is an exact filename.

But, I'll revert that change.

pkg/contexts/ocm/accessmethods/localfsblob/method.go
yitsushi: that was an auto-fixed change with `--fix`. In general if we have more than one `const` or `var` lines, convention to group them together with `const (...)` and `var (...)`, that's why the linter auto-fixed it.

pkg/docker/pusher.go
yitsushi: I see.

This comment, at the top of the file, told me it's "original content":
```
// SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Open Component Model contributors.
//
// SPDX-License-Identifier: Apache-2.0
```

I'll revert the changes and add this directory to the ignore list (with a comment why ignored).

yitsushi: done.

Actions taken:
- reverted all changed under `pkg/docker`
- disabled all checks for `pkg/docker`

cmds/ocm/topics/toi/bootstrapping/topic.go
yitsushi: done.

Actions taken:
- applied suggested changes

pkg/contexts/ocm/accessmethods/localfsblob/method.go
yitsushi: done.

Actions taken:
- applied suggested changes

cmds/ocm/commands/ocmcmds/common/resources.go
mandelsoft: I could also argument exactly the other way round. If I use the input context is several blocks, I always want to use the input context applicable for this block. If I always have to use another variable, depending on the block, the code is in, then this would be very error prone.

But I agree, that in most cases you are right. But as always, it depends on the intended use case.

mandelsoft (approved):

Issue #183, opened 2022-11-09
OCM CLI Documentation Issues/Feedback


In-Ko:
**Description**
The OCM CLI documentation seems to have some smaller issues, which should be addressed:
- [x] OCM itself is designed to be independent of the underlying storage technology. Still, in some parts of the CLI documentation, terms like "registry" or "repository" are used, which are not meaningful in other storage backends than OCI registries. Example: https://ocm.software/docs/cli/download/artefacts/, where it is stated in the description "Download artefacts from an OCI registry."
- [x] The term "Gardener" is mentioned in one of the CLi documentation chapters (here: https://ocm.software/docs/cli/get/credentials/), which is not introduced, so it is not clear what "Gardener" means in this context.
- [x] The examples under https://ocm.software/docs/cli/transfer/artefacts/#examples do not match the command description ("transfer artefacts" is the command vs. "artefact transfer" is mentioned in the examples)
- [x] Docker and OCI is not 100% the same. Still, it is used in the CLI documentation interchangeably.
- [x] The CLI documentation writes "artefact", the Spec "artifact". We must be consistent here and use "artifact" everywhere.

yitsushi:
> OCM itself is designed to be independent of the underlying storage technology. Still, in some parts of the CLI documentation, terms like "registry" or "repository" are used, which are not meaningful in other storage backends than OCI registries. Example: https://ocm.software/docs/cli/download/artefacts/, where it is stated in the description "Download artefacts from an OCI registry."

I understand the outlined issue, but I'm not sure how to resolved it. The CLI documentation does mention OCI and uses that in examples and other places because that's the only fully implemented storage backend in the CLI, alongside local directory/archive file. So yes, OCM itself is not bound to OCI, but the CLI is heavily depends on it.

In-Ko:
Quick comment on this feedback:
> The CLI documentation writes "artefact", the Spec "artifact". We must be consistent here and use "artifact" everywhere.
We decided within the whole team to use artefact, so we will be consistent throughout the documentation and spec, but use british english spelling of the word, which is "artefact".

yitsushi:
> Docker and OCI is not 100% the same. Still, it is used in the CLI documentation interchangeably.

@In-Ko: Can I ask for an example, I used ripgrep to find all `docker` references and it seems all places are correct and not a replacement of "oci".

Examples I found:
> OCI image/artefact reference following the possible docker schemes

> Additionally it provides some limited support for the docker daemon, OCI artefacts and

[...]

PR #199, opened 2022-11-24
refactor: use artifact instead of artefact everywhere

yitsushi:
```bash
rg -l artefact | xargs -I {} sed -i 's/artefact/artifact/g' {}
rg -l Artefact | xargs -I {} sed -i 's/Artefact/Artifact/g' {}

find . -type d -name '*artefact*' | grep -v '.git' | xargs -I {} rename artefact artifact {}
find . -name '*artefact*' | grep -v '.git' | xargs -I {} rename artefact artifact {}
```

Related to: https://github.com/open-component-model/ocm/issues/183

The new renamed version of OCM is now incompatible to the former OCM versions:
- The library packages, types and methods have been changed
- The type names used in the component descriptors have been changed.

All compatibility handling for former OCM versions has been removed.

The OCM toolkit is not yet in use, so we took the chance to get rid of intermediate versions.
Components created with the pre-ocm toolkit should still be valid.


mandelsoft (commented):
Have you added/registered the appropriate additional legacy types for type constants used in existing component versions?

yitsushi:
>  Have you added/registered the appropriate additional legacy types for type constants used in existing component versions?

I'm not sure I understand that. Add/register where?

yitsushi:
Copying content from Slack so it's tracked and others can see it when doing review or looking back in the future.

> There are type constants, like the artifact type ociArtefact which are placed in component descriptors, which are already in some repositories. If such values are just changed, all the existing CVs are corrupted. Therefore we have to
> * introduce new legacy type values for those cases and register them together with the new ones with the appropriate registration methods.
> * Code, which checks an value against such a constant must be adapted to accept both, the old and the new ones. Best, use an appropriate IsXYZ function defined at the place where the constants are defied (instead of checking against the constants directly)
> * There are for sure handlers registered using such constants. They must be registered for the old ones, also. May be, we could use the same mechanism, providing a function returning the values for registration (like TypeSet() []string, then further changed would be easier.
>
> Source: Slack (@mandelsoft)


Can't we ignore as we did not have official release and in that "beta"/"alpha" stage, breaking changes are expected, we don't really have much components yet, and for those we would pollute the system with "legacy" code which sounds extremely  weird on a system that's did not reach 1.0.0 yet (currently `v0.1.0-alpha.1`).

mandelsoft (approved):

PR #202, opened 2022-11-28
docs: change language to liberate ocm from oci specific statements

yitsushi:
I read through all OCI / Docker references, and found only this one. I can't say I 'm 100% sure I found all of them, but the best I can do. All other references contain some degree of separation using words like "for example", or the section is already scoped to `contexts/oci` or `repositories/ocireg`.

Related to: https://github.com/open-component-model/ocm/issues/183

### DIFF
--- a/cmds/ocm/commands/ocicmds/artifacts/download/cmd.go
+++ b/cmds/ocm/commands/ocicmds/artifacts/download/cmd.go
@@ -50,7 +50,7 @@ func (o *Command) ForName(name string) *cobra.Command {
 		Args:  cobra.MinimumNArgs(1),
 		Short: "download oci artifacts",
 		Long: `
-Download artifacts from an OCI registry. The result is stored in
+Download artifacts. The result is stored in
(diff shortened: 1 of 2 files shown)


yitsushi:
The webside can be updated with this job from the ocm repo: https://github.com/open-component-model/ocm-website/actions/workflows/update-cli-docs.yaml

mandelsoft (commented):
Basically the text is correct. This command uses the OCI abstraction to download OCI artifacts
from a repository implementation of the OCI registry abstraction. This is basiaclly
- an OCI registry
- a docker daemon
- a CTF (in OCI interpretation)

This command is not related to the OCM component version abstraction.

yitsushi:
Then as I write under the issue, it's the cli documentation and not ocm documentation. That's why it's under the cli scrope and nothing to change, the cli uses oci, yes ocm is not oci specific but the cli is right now. If it's not specific,we do not mention oci.

PR #215, opened 2022-12-13
get rid of pflag replacement

mandelsoft:
the required new improved API is now replaced by a wrapper method in flag/replace.go, which provides the same behaviour with a more complex implementation, but which is based on the officital API.

Fixes #

**Release note**:

```feature user

```


Skarlso (approved):
Excellent! Great find! :)

PR #235, opened 2023-01-12
add creationTime attribute

mandelsoft:

It adds an optional `creationTimestamp` field to the object meta information.

It uses a format according to [RFC3339](https://www.rfc-editor.org/rfc/rfc3339).

The creation time will be set by the CLI commands creating component versions.
For a library user, the timestamp has to explicitly be added to the component descriptor when
creating a new version.

Correlates to [ocm-spec PR#49](https://github.com/open-component-model/ocm-spec/pull/49).
 
F[ixes #[48](https://github.com/open-component-model/ocm-spec/issues/48)

**Release note**:

```feature user

```


In-Ko:
- Proper documentation of the new element CreationDate is missing (in the OCM and OCM-Spec docs)
- Only one JSON-Schema has been adjusted (the one in the OCM Repo). But there is another one in the OCM-Spec Repo.
  - The OCM-Spec Schema also needs to be changed 
  - This begs the question: Why do we have two separate JSON-Schema anyway ? Is that really necessary ? (If we want do change this, we should do this in another PR though ...)

mandelsoft:
Alignd spec with [PR#49](https://github.com/open-component-model/ocm-spec/pull/49)

In-Ko:
- Discussion about only having 1 JSON Schema is ongoing, will be tackled in a different work package / task.
- With PR https://github.com/open-component-model/ocm-spec/pull/49, the JSON Schema will be updated, so now this PR LGTM.

In-Ko (approved):

In-Ko (approved):
LGTM

PR #263, opened 2023-02-03
Fix expression for RC input

nikimanoledaki:
Fixing the expression for the Release Candidate input in the conditional that sets the version.


phoban01 (approved):
LGTM. Thanks for catching.

nikimanoledaki:
@phoban01 you were right about this one :D

PR #267, opened 2023-02-16
check type assertions where it makes sense to avoid runtime panics

### LINKED ISSUE #22: Over all `nil` handling in the code when doing type assertions and validations is missing
If there is a slight error, or a missing type or a type assertion is the type that you want, or there is a slight error in anything, it leads to a crash.

```
ocm sign componentversions test-ocm -s skarlso-sig -K private-key.pem -k public-key.pem
applying to version "skarlso.test/name:0.0.1"...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x123d234]

goroutine 1 [running]:
github.com/open-component-model/ocm/pkg/common.VersionedElementKey({0x0?, 0x0})
        /Users/skarlso/goprojects/SAP/ocm/pkg/common/types.go:46 +0x54
github.com/open-component-model/ocm/pkg/contexts/ocm/signing.apply({0x26804c0, 0xc0009e5ef0}, {0xc00099f8c0, {0xc000634460, 0x1, 0x1}}, {0x0?, 0x0}, 0xc000756790)
        /Users/skarlso/goprojects/SAP/ocm/pkg/contexts/ocm/signing/handle.go:49 +0xaa
github.com/open-component-model/ocm/pkg/contexts/ocm/signing.apply({0x26804c0, 0xc00099f890}, {0xc00099f8c0, {0xc000634460, 0x1, 0x1}}, {0x2697fa8?, 0xc0009ddd28}, 0xc00091fab0)
        /Users/skarlso/goprojects/SAP/ocm/pkg/contexts/ocm/signing/handle.go:72 +0x33b8
github.com/open-component-model/ocm/pkg/contexts/ocm/signing.Apply({0x26804c0?, 0xc00099f890?}, 0x17b1c54?, {0x2697fa8, 0xc0009ddd28}, 0x2697fa8?)
        /Users/skarlso/goprojects/SAP/ocm/pkg/contexts/ocm/signing/handle.go:45 +0x13b
github.com/open-component-model/ocm/cmds/ocm/commands/ocmcmds/common/cmds/signing.(*action).Add(0xc0002a4b60, {0x2252ee0?, 0xc000258840})
        /Users/skarlso/goprojects/SAP/ocm/cmds/ocm/commands/ocmcmds/common/cmds/signing/cmd.go:134 +0x33a
github.com/open-component-model/ocm/cmds/ocm/pkg/utils.HandleOutput({0x2680160, 0xc0002a4b60}, {0x2680190, 0xc00099f7d0}, {0xc0004f2c30?, 0x1, 0xc00091fc78?})
        /Users/skarlso/goprojects/SAP/ocm/cmds/ocm/pkg/utils/handling.go:102 +0x282
github.com/open-component-model/ocm/cmds/ocm/commands/ocmcmds/common/cmds/signing.(*SignatureCommand).Run(0xc00029c000)
        /Users/skarlso/goprojects/SAP/ocm/cmds/ocm/commands/ocmcmds/common/cmds/signing/cmd.go:100 +0x4f5
github.com/open-component-model/ocm/cmds/ocm/pkg/utils.SetupCommand.func1(0xc000278500?, {0xc0002a4930, 0x1, 0x7})
        /Users/skarlso/goprojects/SAP/ocm/cmds/ocm/pkg/utils/command.go:103 +0xff
github.com/spf13/cobra.(*Command).execute(0xc000278500, {0xc0001a8030, 0x7, 0x7})
        /Users/skarlso/go/pkg/mod/github.com/mandelsoft/cobra@v1.4.1-0.20220602090819-209e4a9a442d/command.go:874 +0x694
github.com/spf13/cobra.(*Command).ExecuteC(0xc000457400)
        /Users/skarlso/go/pkg/mod/github.com/mandelsoft/cobra@v1.4.1-0.20220602090819-209e4a9a442d/command.go:992 +0x3b4
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/skarlso/go/pkg/mod/github.com/mandelsoft/cobra@v1.4.1-0.20220602090819-209e4a9a442d/command.go:920
main.main()
        /Users/skarlso/goprojects/SAP/ocm/cmds/ocm/main.go:27 +0x2c
```

This is coming from the library, which when used by a controller could result in a catastrophic crash.

The code must be strengthened and nil checks for all types of reference-based access must be added.

On top of that, it's not a nice user-experience, since now, I have to dig around in the code to find out what went wrong. :)

jmickey:

#22 - This PR adds type assertion checks where it feel sensible in order to avoid `nil` pointer referencing that causes runtime panics. 

There are still a lot of type assertions occurring in the code, but adding checks for many of these either didn't make sense in the context, or would require code refactoring beyond the scope of the ticket.

Fixes #22

### DIFF
--- a/cmds/ocm/commands/ocicmds/artifacts/transfer/cmd.go
+++ b/cmds/ocm/commands/ocicmds/artifacts/transfer/cmd.go
@@ -152,7 +152,10 @@ func NewAction(ctx clictx.Context, session oci.Session, target string, transferR
 }
 
 func (a *action) Add(e interface{}) error {
-	src := e.(*artifacthdlr.Object)
+	src, ok := e.(*artifacthdlr.Object)
+	if !ok {
+		return fmt.Errorf("failed type assertion for type %T to artifacthtlr.Object", e)
+	}
 
 	ns := src.Namespace.GetNamespace()
 	if ns == "" && a.Ref.IsRegistry() {
--- a/pkg/contexts/credentials/repositories/memory/config/type.go
+++ b/pkg/contexts/credentials/repositories/memory/config/type.go
@@ -83,7 +83,10 @@ func (a *Config) ApplyTo(ctx cfgcpi.Context, target interface{}) error {
 		return fmt.Errorf("unable to get repository for spec: %w", err)
 	}
 
-	mem := repo.(*memory.Repository)
+	mem, ok := repo.(*memory.Repository)
+	if !ok {
+		return fmt.Errorf("invalid type assertion of type %T to memory.Repository", repo)
+	}
 
 	for i, e := range a.Credentials {
 		var creds cpi.Credentials
--- a/cmds/ocm/commands/ocmcmds/common/addhdlrs/comp/elements.go
+++ b/cmds/ocm/commands/ocmcmds/common/addhdlrs/comp/elements.go
@@ -67,7 +67,7 @@ func (*ResourceSpecHandler) Add(ctx clictx.Context, ictx inputs.Context, elem ad
 
 	r, ok := elem.Spec().(*ResourceSpec)
 	if !ok {
-		return fmt.Errorf("element spec is not a valid resource spec")
+		return fmt.Errorf("element spec is not a valid resource spec, failed to assert type %T to ResourceSpec", elem.Spec())
 	}
 	comp, err := repo.LookupComponent(r.Name)
 	if err != nil {
--- a/cmds/ocm/commands/ocmcmds/common/addhdlrs/comp/elements.go
+++ b/cmds/ocm/commands/ocmcmds/common/addhdlrs/comp/elements.go
@@ -65,7 +65,10 @@ func (*ResourceSpecHandler) Add(ctx clictx.Context, ictx inputs.Context, elem ad
 	var final utils.Finalizer
 	defer final.FinalizeWithErrorPropagation(&err)
 
-	r := elem.Spec().(*ResourceSpec)
+	r, ok := elem.Spec().(*ResourceSpec)
+	if !ok {
+		return fmt.Errorf("element spec is not a valid resource spec")
+	}
 	comp, err := repo.LookupComponent(r.Name)
 	if err != nil {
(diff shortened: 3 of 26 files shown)


cmds/ocm/commands/ocicmds/artifacts/transfer/cmd.go
```diff
@@ -152,7 +152,10 @@ func NewAction(ctx clictx.Context, session oci.Session, target string, transferR
 }
 
 func (a *action) Add(e interface{}) error {
-	src := e.(*artifacthdlr.Object)
+	src, ok := e.(*artifacthdlr.Object)
+	if !ok {
+		return fmt.Errorf("failed type assertion for type %T to artifacthtlr.Object", e)
```
mandelsoft, line 157: Basically this would be a local programming error. which is directly detected by the test cases, it is not a library function. So, I'm not sure, whether these runtime checks are really required. But anyway, the other locations should use this `%T`, also, to get a glimpse on what went wrong.

pkg/contexts/credentials/repositories/memory/config/type.go
```diff
@@ -83,7 +83,10 @@ func (a *Config) ApplyTo(ctx cfgcpi.Context, target interface{}) error {
 		return fmt.Errorf("unable to get repository for spec: %w", err)
 	}
 
-	mem := repo.(*memory.Repository)
+	mem, ok := repo.(*memory.Repository)
```
mandelsoft, line 86: Even here, the flow explicitly creates a memory.Repository by a generic function getting a spec for such a repository.
So, basically it must be of this type.

But you are right, these are typical *famous last words* :-)

jmickey:
@mandelsoft certainly understand where you're coming from. I think, even if we don't expect there to be issues caused by these type assertions, that checking them is still a worthwhile exercise. It adds very little code overall, and we remove an unknown variable when encountering possible runtime panics. Effectively, I don't think it hurts _to be sure_ that these assertions won't cause us issues moving forward.

Skarlso (approved):
LGTM

PR #273, opened 2023-02-21
#105 - Panic Handlers

### LINKED ISSUE #105: panic ADR research: implement a handling POC based on the ADR

Implement, or research what it takes to implement, the panic ADR located [here](https://github.com/gardener/ocm/blob/main/docs/adrs/0002-panic-handlers.md).

**Why is this needed**:

jmickey:

Allow lib callers to register panic handlers, and choose decide if panics will continue to cause the code to crash.

A defer function has been added to the call sites where panics are made in the library code. Initially I was trying to follow the references up the stack to see if there were fewer places where a defer could be made, but in the end it made the more sense to place the defer close to the panic itself.

Lib consumers can register their own handlers that matches the `PanicHandler` type - `func(interface{})`:

```go
import "github.com/open-component-model/ocm/pkg/util/panics"

func main() {

	panics.RegisterPanicHandler(func (in interface{}) {
		// do something
		return true
	})

}
```

By default panics will always log the stacktrace.

Fixes #105


Skarlso (dismissed):
Nice work! :)

mandelsoft (changes requested):
I'm wondering how (and why) you have selected the locations to place the `defer`. It looks like you have chosen places near the occurrence of panics. But this typically leads to a continuing program flow with undefined results, where the caller has no possibility to identify this case, which basically will lead to further runtime panics. These panics are for sure NOT handled by a panic handler, again, so that you still get panics, but now completely unrelated to the original problem.

Similar to the K8S handling, the `defer` should be placed near a resurrection point in a program flow, where typically a loop for request handling is in place to assure the handling of further requests, even if a single single one runs in such a problematic situation (for example, reconcilation loop). Or at least a place, where you can safely ignore the outcome of a called function and continue with a new calculation context.

A defer inside an ìnit function is questionable. If the program initialization fails because of some basic misalignment, it makes no sense to run the process at all.

The `Must` versions of functions panic if the non-Must versions return an error.
Placing `defers` here is also questionable, because you could directly call the error providing function if you want to deal with failures. The Must version is typically intended to be called in init functions (see above).

jmickey:
> I'm wondering how (and why) you have selected the locations to place the defer. It looks like you have chosen places near the occurrence of panics.

Correct. Restricted to panics in the library code.

> But this typically leads to a continuing program flow with undefined results, where the caller has no possibility to identify this case, which basically will lead to further runtime panics. These panics are for sure NOT handled by a panic handler, again, so that you still get panics, but now completely unrelated to the original problem.

I would argue that this isn't our concern, but rather the users concern. Optimally we wouldn't `panic` from library code, but because we do, we should at least provide the option to handle those panics as the user sees fit. I would imagine that in the vast majority of cases the crash will continue. If a user opts to continue the program flow then the consequences of that are for them to deal with.

> Similar to the K8S handling, the defer should be placed near a resurrection point in a program flow, where typically a loop for request handling is in place to assure the handling of further requests, even if a single single one runs in such a problematic situation (for example, reconcilation loop). Or at least a place, where you can safely ignore the outcome of a called function and continue with a new calculation context.

I'm not really sure I understand what your saying here, and how it relates in the context of OCM and `defer` placements.

> A defer inside an ìnit function is questionable. If the program initialization fails because of some basic misalignment, it makes no sense to run the process at all.

If you feel strongly about this I can remove the `defer`s from inits. Panicking from an init is, at least to me, also questionable unless that `init` sits in code that is wholly owned by an executable context rather than lib context (e.g. a cmd executable).

> The Must versions of functions panic if the non-Must versions return an error.

The user still may wish to execute some process or code before continuing with a crash, so I think it is still valid to call the handlers here?

Skarlso:
We'll do a compromise and place handling at certain points only. The rest will be handled in a different manner.

The ADR has been accepted by all parties so we'll try to be more in line with that. :)

In-Ko:
@jmickey / @Skarlso / @mandelsoft : Can we please get this clarified this week ? It is a long runner already, and I'd like to see this being finished soon. My take: If the current implementation is done according to the ACD which we all agreed on, than this should go out.

Skarlso:
Right, so the main points are:

- The problem with catching a panic where they are is that somewhere along the line another panic will happen because now the result is undefined or even nil so the resulting error will be unrelated to the first panic causing problem.
- Put the handler into a place of recovering.
Regarding this point.... There are no safe places in OCM after a panic. There are lot of places where error handling is non-existent so this is at least not really possible without significant changes. I guess in those cases, we will just ignore the panic. I don't have an alternative.
- Remove the defer in init
- Ignore MUST functions.

pkg/signing/handlers/rsa/format.go
```diff
@@ -67,6 +68,9 @@ func KeyData(key interface{}) ([]byte, error) {
 }
 
 func PemBlockForKey(priv interface{}, gen ...bool) *pem.Block {
+	// Handling the panic here means that the above write will error and return that error which might be
+	// acceptable by the end user.
+	defer panics.HandlePanic()
```
mandelsoft, line 73: What happend here: in the body there is a panic because a detected inconsistent state.
In the same function there is panic defer which would result in a nil return.

Why not replace the panic directly by a nil return?
What happens with a nil return? The caller expectes to get a pem block, so it will probaply cause in implicit panic because of a nil pointer access up the call tree.

What should be the advantage? There is finally still a panic, but now you don't know the reason anymore.

A panic defer in the same function providing a panic make really no sense. 
- If there is a nil-able return value indicating an invalid state, then we could change the contract and make returning nil a part of it.
- If there is an error return we should return an error.
- If there is no return at all, then it makes definitely no sense to ignore the panic because it directly leads to inconsistent behaviour up the call tree.
  
If at all the panic should be catched at a place where:
- either the call tree can safely be continued (for example a request can just be aborted and continued with the next request)
- the function provides an error. Then we should return the panic as error. Therefore we would need the possbility to convert the panic into an error return. 

Therefore we need a flavor of the defer function taken an error pointer, which is fed by the recovered error in case the panic should be catched.

All other locations make definitely no sense to place such a defer, because it typically leads to worse error behavior.

pkg/utils/panics/panics.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/open-component-model/ocm/pkg/logging"
+)
+
+var ReallyCrash = true
+
+type PanicHandler func(interface{})
+
+var PanicHandlers = []PanicHandler{logHandler}
+
+func HandlePanic(additionalHandlers ...PanicHandler) {
```
mandelsoft, line 20: Here we should also provide an flavor able to forward the panic as an error return (see above)

pkg/contexts/ocm/utils.go
```diff
@@ -35,7 +35,7 @@ func AssureTargetRepository(session Session, ctx Context, targetref string, opts
 		case string:
 			archive = v
 		default:
-			panic(fmt.Sprintf("invalid option type %T", o))
+			return nil, fmt.Errorf("invalid option type %T", o)
```
mandelsoft, line 38: yep, this is much better, in  error returning functions a panic is not applicable at all.

pkg/cobrautils/links.go
```diff
@@ -32,6 +34,8 @@ func FormatLinkWithHandler(linkhandler func(string) string) func(string) string
 }
 
 func SubstituteCommandLinks(desc string, linkformat func(string) string) ([]string, string) {
+	// This is acceptable as the call chain will just output empty.
+	defer panics.HandlePanic()
```
mandelsoft, line 38: Basically, this panic here is to inform the developer, that there is an error its command description. It is only used for CLIs If the CLI does not enable the ignore panic mode, there would be no change in behavior. And this is not enabled by the CLI (I hope).

Silently ignoring this would result in undetected corrupted command documentation.

cmds/ocm/pkg/utils/handling.go
```diff
@@ -45,6 +46,8 @@ func StringElemSpecs(args ...string) []ElemSpec {
 }
 
 func ElemSpecs(list interface{}) []ElemSpec {
+	// This is acceptable as the calling chain will just take a nil slice.
+	defer panics.HandlePanic()
 	if list == nil {
```
mandelsoft, line 51: see comment below concerning panic defers in functions with a panic.

Basically we should catch this problem up the call tree, where the processing chain is executed and an error return is possible. At that place it should be converted to an error,

A good place would be the initial main command execution, the command is aborted because of a severe programming error, and this error is returned to the use as regular error. A panic defer at that place would be a good idea anyway to prevent callstacks to reach the CLI user. But then we would need an option, to enable the callstack again, to give the developer a chance to figure out the concrete location.

mandelsoft (changes requested):
Good job, to find all the places where a regular error return can do the job to avoid panics at all.

One panic should be avoided by refactoring the interface and add an error return value.
If you want I can do this at short notice. The are probably many places involved.

pkg/signing/handlers/rsa/format.go
Skarlso: The panic handler would help in logging the issue. A handler here would add some context to the nil writer. Either that or we add something by logging it out. What do you think? The reason for a handler here in my thought is that, for example, a handler could include a prometheus aggregator that tracks errors by emitting some metric.

>What should be the advantage? There is finally still a panic, but now you don't know the reason anymore.

Sure, but the handler supposedly logged the upward panic already. So ideally, there would be two lines of logs catching both panics. That said, I thought I removed all downwards panics, I might have missed this one. :)

Skarlso: Remeber that the caller might still throw the panic, they just want some kind of extra information attached to it.

pkg/contexts/oci/repositories/ocireg/namespace.go
```diff
@@ -163,18 +175,25 @@ func (n *NamespaceContainer) AddArtifact(artifact cpi.Artifact, tags ...string)
...
 	}
+
 	n.repo.ctx.Logger().Debug("adding artifact", "digest", blob.Digest(), "mimetype", blob.MimeType())
-	_, _, err = n.blobs.Get(blob.MimeType()).AddBlob(blob)
+	blobData := n.blobs.Get(blob.MimeType())
+	if blobData == nil {
+		return nil, fmt.Errorf("failed to retrieve blob data, blob data was empty")
+	}
+
+	_, _, err = blobData.AddBlob(blob)
```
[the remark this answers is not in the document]
Skarlso: Yes, that makes sense. Good idea.

pkg/contexts/oci/repositories/ocireg/blobs.go
```diff
@@ -86,7 +86,8 @@ func NewBlobContainer(cache accessio.BlobCache, mime string, fetcher resolve.Fet
 	}
 	r, err := accessio.CachedAccess(c, c, cache)
 	if err != nil {
-		panic(err)
+		// The calling code chain handles this blob container being nil.
+		return nil
```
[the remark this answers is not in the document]
Skarlso: 👍 That's a good idea. I like that. :)

pkg/cobrautils/links.go
Skarlso: Sounds good. If this is not library code, the panic is acceptable.

cmds/ocm/pkg/utils/handling.go
Skarlso: But the initial command execution is too far above, right? If this util function could be called from the outside by something that imports OCM we should protect that calling library. Not?

pkg/utils/panics/panics.go
Skarlso: Just be aware though that in that case, we'll have to wrap the returning value. Because the Handler needs to be called in a defer.

cmds/ocm/pkg/processing/buffer.go
```diff
@@ -365,6 +369,9 @@ func (ob *orderedBuffer) SetFrame(frame BufferFrame) {
 }
 
 func (ob *orderedBuffer) Add(e ProcessingEntry) bool {
+	// This handler here prevents the calling function's lock to be not released.
```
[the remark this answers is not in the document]
Skarlso: Sure. But that is super difficult in some of the cases because the functions are called either recursively or through many indirections. It's impossible to tell where the snake begins. :)

This particular function appears to be called from the parallel executor if I see it correctly. Which is called from a bunch of things that I was maybe able to track back to some cli setup. So where do you think would be a good place to catch this one? :)

cmds/ocm/pkg/utils/handling.go
Skarlso: Actually, I can catch this in the Command error Run.

pkg/utils/panics/panics_test.go
```diff
@@ -0,0 +1,63 @@
...
+			Expect(recover()).To(BeNil())
+		}()
+		err := caller("exception", nil)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(MatchRegexp(`(?s)Observed a panic: "exception"
+goroutine [0-9]* \[running\]:
+panic.*$`))
+	})
+
+	FIt("propagates catched panic with handlers", func() {
```
Skarlso, line 42: don't forget to remove this `FIt` :)

pkg/utils/panics/panics.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+}
+
+// RegisterPanicHandler adds handlers to the panic handler.
+func RegisterPanicHandler(handler PanicHandler) {
+	PanicHandlers = append(PanicHandlers, handler)
+}
+
+func mapRecovered(r interface{}, offset int) error {
```
Skarlso, line 69: What is the purpose of `offset` and the append lines? 🤔 That is way too much magic. :) I'm sure there is a simpler way of handling that Join.

Skarlso:
@mandelsoft This looks good to me now. :) Thanks for the additions. 👍

pkg/signing/handlers/rsa/format.go
mandelsoft, line 73: Here, we can use an error return, also
I'll fix this.

mandelsoft (changes requested):

PR #283, opened 2023-03-02
support NPM registry access

mandelsoft:

Add new access method for accessing NPM packages in NPM registries.
This is just a happy path implementation.  The registry spec is quite superficial. 
 I couldn't find something about authentication, so this is not implemented.

It provides sha1 hash to validate the content. This validation is implemented.
Then there is an integrity field not described by the spec containing a SHA512 hash.
I don't know, whether is always supported. It is base64 encoded which does not match the regular
digest format and would require additional conversion.

Fixes #

**Release Notes**:

```feature user

```


morri-son (approved):
looks ok

PR #305, opened 2023-03-15
support of sub charts for toi helm installer

mandelsoft:

The helm TOI installer config now allows to specify additional helm chart resources to be used as nested helm charts,
Fixes #

**Release Notes**:

```feature user

```


jensh007 (approved):

jensh007 (changes requested):
INFO[0002] validating executor config by scheme...      
Error: error executor config: executor config validation failed: (root): Additional property subcharts is not allowed

The executorspec.yaml needs a schema update.

cmds/helminstaller/README.md
```diff
@@ -0,0 +1,52 @@
...
+It provides a single `toiExecutor` resource with the
+name `toiexecutor`.
+
+## Configuration
+
+The executor configuration supports the following field:
+
+- `chart` *[ResourceReference](../../docs/reference/ocm_toi-bootstrapping.md#resourcereference)* (**required**) The resource containing the top level hemlm chart to work on.
+
+- `subCharts` *map[string]ResourceReference* (**optional**) A set of resources describing charts used as sub charts for the top level chart.
```
jensh007, line 21: Documentation is insufficient here:

- What is the meaning of the key in this map or why is it not an array of ResourceReferences?
Example
```
        foo:
          resource:
            name: nginx-controller-chart
          referencePath:
          - name: nginx-controller
        bar:
          resource:
            name: echo-chart
          referencePath:
          - name: echoserver
```
What is the meaning of foo/bar?

What happens with imageMapping for subcharts?

Example would be helpful.

mandelsoft: It uses the value propagation mechanism of the helm subchart mechanism. See https://helm.sh/docs/chart_template_guide/subcharts_and_globals/. So, there is nothing special.
I've added a short description to the README.

The key of the subchart map is the name of the subchart folder below the parent chart's `charts` folder.
I also added this to the description.

jensh007 (approved):

jensh007 (approved):

jensh007 (approved):

PR #307, opened 2023-03-20
simple data localization

### LINKED ISSUE #300: Add substitution capability that doesn't require a virtual filesystem or a filename

Ability to perform substitution on plain byte data.

**Why is this needed**:

The substitution functionality right now requires a VFS and files to be passed in. But in same cases, when we only have a blob, we can't pass in a filename. And since libraries aren't aware of Acces methods ( which are raw JSON ), we can't have a `path` information to fetch the name of the file to create a tar file ourselves.

mandelsoft:
This PR deals with localization on a single data object.

The change introduces some basic functions around a type `ValueMappings` , which just describes a set of substitution rules, which can be applied to a single data object (`SubstituteMappings`) . With `LocalizeMappings` a set of (pure non-filename based) localization rules can be mapped to such value mappings combined with config mappings and then applied to a single byte array,

Fixes #300

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/utils/localize/README.md
+++ b/pkg/contexts/ocm/utils/localize/README.md
@@ -48,3 +48,7 @@ Such a specification object can be applied by the function `Instantiate`
 together with configuration values to
 a component version. As substitution result it returns a virtual filesystem
 with the snapshot according to the resolved substitutions.
+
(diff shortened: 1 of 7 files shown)


Skarlso (approved):
Nice one! Thanks for this.

PR #324, opened 2023-04-04
Adjust dockerconfig.RepositorySpec 

fabianburth:
This PR extends the dockerconfig.RepositorySpec to also being able to deal with the docker config data directly and not only with docker config files within a filesystem. This enables to use this for docker config data which was e.g. retrieved from kubernetes secrets. 

**Release Notes**:


pkg/contexts/credentials/repositories/dockerconfig/repo_test.go
```diff
@@ -51,18 +53,57 @@ var _ = Describe("docker config", func() {
...
+			datajson map[string]interface{}
+			specjson map[string]interface{}
+		)
+		// Comparing the bytes might be problematic as the order of the JSON objects within the config file might change
+		// during Marshaling
+		err = json.Unmarshal([]byte(specdata3), &specjson)
+		Expect(err).To(Succeed())
+		err = json.Unmarshal(data, &datajson)
+		Expect(err).To(Succeed())
+		Expect(datajson).To(Equal(specjson))
```
mandelsoft, line 73: The additional code de-facto describes separate test cases, so this should be done by  separate test cases.

pkg/contexts/credentials/repositories/dockerconfig/type.go
```diff
@@ -24,8 +25,9 @@ func init() {
 // RepositorySpec describes a docker config based credential repository interface.
 type RepositorySpec struct {
 	runtime.ObjectVersionedType `json:",inline"`
-	DockerConfigFile            string `json:"dockerConfigFile"`
-	PropgateConsumerIdentity    bool   `json:"propagateConsumerIdentity,omitempty"`
+	DockerConfigFile            string          `json:"dockerConfigFile"`
```
mandelsoft, line 28: Here an `omitempty` would be useful, because it is now optional.

mandelsoft (changes requested):

mandelsoft (approved):

PR #326, opened 2023-04-05
cleanup OCI layers in OCM mapping

### LINKED ISSUE #323: When uploading a ctf for an existing component again, the artifact is not overwritten, but a new blob is added
**What happened**:
existing artifact not replaced, but additional one added
existing artifact should be replaced
```ocm add cv```with content like 
```resources:
  - name: blueprint
    type: blueprint
    relation: local
    input:
      type: dir
      path: ./blueprint
      mediaType: "application/vnd.gardener.landscaper.blueprint.v1+tar+gzip"
      compress: true
```
and transfer the ctf multiple times and then check the artifacts of the cv.

**Environment**:

mandelsoft:

If a component version is updated in an OCI based component respository,
new local blob content is just added as new layers, but if an already existing local resource
is modified, then old layer holding the old blob content is not removed again.

As a result the OCI artifact is growing with every update. 

This has no influence on the OCM functionality, even transports are not affected,
but on the OCI side the artifact is growing.

This PR now identifies the required layers and removes the obsolete ones again from the layer list of the OCI
artifact when a ComponentVersionAccess is updated on the underlying OCI storage backend.

Unfortunately with some legacy CNUDIE access method (localOCIBlob) this is not always possible, because
the MediaType is not stored in the access specification, but in the OCI manifest, only. If the same content 
is used by multiple OCM resources, but with different MediaTypes, always the oldest layer is deleted.
But this was basically a problem also for the legacy implementation.

Fixes #323

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/compdesc/componentdescriptor.go
+++ b/pkg/contexts/ocm/compdesc/componentdescriptor.go
@@ -573,6 +573,14 @@ func (o *ResourceMeta) Copy() *ResourceMeta {
 	return r
 }
 
+func NewResourceMeta(name string, typ string, relation metav1.ResourceRelation) *ResourceMeta {
+	return &ResourceMeta{
+		ElementMeta: ElementMeta{Name: name},
+		Type:        typ,
+		Relation:    relation,
+	}
+}
+
 type References []ComponentReference
 
 func (r References) Len() int {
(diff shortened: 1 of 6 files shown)


jensh007 (approved):

morri-son (approved):
tested and looks ok

PR #330, opened 2023-04-08
Support Consumer Providers + Migrate DockerConfig

mandelsoft:

Instead of propagating single credentials for single consumers, it is possible now
to add complete CosumerProviders handling consumer requests.

The explicit consumer registry has been replaced by an internal standard consumer provider
and the Docker Config repository has been migrated to provide such a consumer provider.

Consumer providers and entries now have an optional provider id, which can be used
to clean up all setting from one provider. This can be used to refresh or delete providers,
again.

Additionally, the contexts now feature a unique Id for debugging and logging.
They also offer a possibility now to check, whether a context has been garbage collected
as expected when getting rid of a context.

The Credentials Context now has a test for this.

Fixes #

**Release Notes**:

```feature user

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 
 require (
 	github.com/Masterminds/semver/v3 v3.2.0
+	github.com/aws/aws-sdk-go v1.15.11
 	github.com/aws/aws-sdk-go-v2 v1.17.7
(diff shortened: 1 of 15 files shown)


jensh007 (dismissed):

mandelsoft:
had to rebase

PR #348, opened 2023-04-24
use OCMBot token for event publishing

mandelsoft:

Fixes #

**Release Notes**:

```feature user

```


mandelsoft:
@phoban01, so far, the release event is only published for non-pre releases. Do we want to stick to this?

phoban01 (approved):

PR #349, opened 2023-04-24
generalize resource selectors

mandelsoft:

Offer a public resource selector feature on the internal component descriptor version.

Fixes #

**Release Notes**:

```feature user

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ The library currently supports the following [repository mappings](docs/ocm/inte
 - **Component Archive**: Compose the content of a component version on the
   filesystem.
 
+For the usage of the library to access OCM repositories, handle configuratio and credentials see the [examples section](examples/lib/README.md).
+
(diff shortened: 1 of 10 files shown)


phoban01 (dismissed):

Skarlso (dismissed):
Nice! Thanks, Uwe!

mandelsoft:
I've added resource selection via Labels and additionally tests for the new features.

Skarlso:
Please try and not extend this any further. :D Also, we kind of need this for our tests sooner rather than later. :)

thanks! :)

mandelsoft:
yes, it is done now. What I wanted to achieve are the same selection features for resources and references (and potentially sources).
Therefore, the selector interface had to be extended a little bit. But now it is in a shape to be extendable in a compatible manner. if some extension will be required in the future.

jensh007 (approved):

PR #355, opened 2023-05-04
OCI tag mapping

mandelsoft:
Unfortunately we have little problem with our OCM to OCI mapping:
So far, we just use OCM version names as tags for OCI manifests. And OCM version names must follow the semantic versioning scheme. This scheme uses the `+` character to append a build/metadata specification part. Unfortunately this character is forbidden in an OCI tag name according to the OCI distribution spec.

There are discussions about this since at least 2017, which finally fail to come to a solution (especially because + is a reserved character in URLs and should not be used there)

There are OCI implementations (like artifactory) allowing such tags (but the OCM lib cannot get artifacts with those tags), others like [ghcr.io](http://ghcr.io/) and [gcr.io](http://gcr.io/)  reject such tags.
Nevertheless we want to stick to the semver scheme for our OCM component version names.
Therefore, we have to define a version-to-tag mapping for the OCM to OCI mapping specification.

There are several ways to do this, with pros and cons:
- mostly unique mapping, but incompatible with actually used versions: replace `-` by `--` and `+` by `-` in the pre-release part of the version. Even this has the problem with mapping something like `---` back to a version.
- non-unique, but compatible mapping: replace `+` in the pre-release part by an unlikely sequence like `-.-` or `-plus-`.

Any preferences? Or may be even better ideas?

The actual implementation uses `.build-` to substitute the metadata data separator char (`+`). If keeps the pre-release parts (separated by dots) unchanged and just adds the metadata suffix as an additional and last pre-release part prefixed with `build-` to indicate the meaning as metadata part,

Additionally, the PR fixes the ocm ref parsing to fully support the complete semantic version syntax. Before, only the OCI version syntax was accepted. Therefore specifying a semantic version with a metadata path was not accepted (or misinterpreted).

Fixes #

**Release Notes**:

```feature user

```


docs/ocm/modelmapping/oci/README.md
```diff
@@ -48,6 +48,18 @@ access specification.
 
 </div>
 
+## Version Mapping
+
+The Open Component Model supports version names according to [semantic versioning](https://semver.org/).
+The tags used to represent versions in the OCI specification do not allow to directly use semantic version names as tags, becase the plus (`+`) character is not supported. Therefore, the open component model version names have to be mapped
+to OCI-compliant tag names.
+
+The followinmg mapping for version is used, here:
```
ccwienk, line 57: typo: followinmg -> following

docs/ocm/modelmapping/oci/README.md
```diff
@@ -48,6 +48,18 @@ access specification.
 
 </div>
 
+## Version Mapping
+
+The Open Component Model supports version names according to [semantic versioning](https://semver.org/).
+The tags used to represent versions in the OCI specification do not allow to directly use semantic version names as tags, becase the plus (`+`) character is not supported. Therefore, the open component model version names have to be mapped
```
ccwienk, line 54: typo: becase -> because

docs/ocm/modelmapping/oci/README.md
```diff
@@ -48,6 +48,18 @@ access specification.
 
 </div>
 
+## Version Mapping
+
+The Open Component Model supports version names according to [semantic versioning](https://semver.org/).
```
ccwienk, line 53: nit: since we allow `v` prefix, we are not strictly semver-compliant

jensh007 (dismissed):

docs/ocm/modelmapping/oci/README.md
mandelsoft: added a footnote

jensh007 (approved):

PR #364, opened 2023-05-10
Consistent handling of spec versions

mandelsoft:

Changes on the CPI-level, only

- rework the handling of different spec versions for specification kinds (for example RepositorySpecs)
- provide version support for all kinds of specifications for all context types
- redesign of embedding of groups of OCM repository specs based on code (used for supporting all OCI repo types as OCM repo type). Be prepared for multiple such groups (for example for an BlobStore content type).

- first steps towards generic runtime types.

Fixes #

**Release Notes**:

```feature user

```


pkg/runtime/convert.go
```diff
@@ -0,0 +1,80 @@
...
+func (c *ConvertWrapper[T]) ConvertTo(object interface{}) (TypedObject, error) {
+	return c.converter.ConvertTo(object)
+}
+
+func NewConvertWrapper[T TypedObject](c Converter[T]) TypedObjectConverter {
+	return &ConvertWrapper[T]{converter: c}
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
```
fabianburth, line 27: FormatVersion, ConvertedType  and VersionedType should be unified (in a follow-up PR). 
If the external format version is equal to the internal format version, the conversion should be the identity. 

Such a change should reduce the number and nesting of interfaces.

fabianburth (changes requested):

fabianburth (approved):

PR #369, opened 2023-05-17
time based cache cleanup

mandelsoft:

The filesystem cache for blobs now supports an API for a cache cleanup.
It supports a dry-run mode and a time based mode.

It is accessible using an option interface `CacheCleanup`.

This way it can programatically be used from a library without the command line interface.,

Fixes #

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/common/accessio/cache.go
+++ b/pkg/common/accessio/cache.go
@@ -160,6 +160,12 @@ var (
 	_ RootedCache = (*blobCache)(nil)
 )
 
+// ACCESS_SUFFIX is the suffix of an additional blob related
+// file used to track the last access time by its modification time,
+// because Go does not support a platform independent way to access the
+// last access time attribute of a filesystem.
+const ACCESS_SUFFIX = ".acc"
+
 func NewDefaultBlobCache(fss ...vfs.FileSystem) (BlobCache, error) {
 	var err error
 	fs := DefaultedFileSystem(nil, fss...)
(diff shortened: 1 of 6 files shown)


pkg/common/accessio/cache.go
```diff
@@ -204,6 +270,9 @@ func (c *blobCache) GetBlobData(digest digest.Digest) (int64, DataAccess, error)
 		path := common.DigestToFileName(digest)
 		fi, err := c.cache.Stat(path)
 		if err == nil {
+			vfs.WriteFile(c.cache, path+".acc", []byte{}, 0o600)
```
fabianburth, line 273: consider using constant instead of literal for ".acc"

pkg/common/accessio/cache.go
```diff
@@ -272,6 +341,7 @@ func (c *blobCache) AddBlob(blob BlobAccess) (int64, digest.Digest, error) {
 		err = c.cache.Rename(tmp, target)
 	}
 	c.cache.Remove(tmp)
+	vfs.WriteFile(c.cache, target+".acc", []byte{}, 0o600)
```
fabianburth, line 344: replace ".acc" with above mentioned constant

pkg/utils/time.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"strconv"
+	"time"
+
+	"github.com/open-component-model/ocm/pkg/errors"
+	"github.com/open-component-model/ocm/pkg/generics"
+)
+
+func ParseDeltaTime(s string, past bool) (time.Time, error) {
```
fabianburth, line 15: perhaps add a short documentation, as the parameters do not make it obvious how this function is supposed to be used

fabianburth (changes requested):

fabianburth (approved):

PR #372, opened 2023-05-24
fix default github token handling + compatibility handling for oci

mandelsoft:

-for GitHub, token based auth always requires an arbitrary user name.
- modifying CV using a legacy format always updated the mimetypes. With this fix
  the original mime types are kept
- the compatattr is now used to control the mimetypes used for the OCi mapping of OCM
- bump  spiff to a released version

Fixes #

**Release Notes**:

```feature user

```


morri-son (commented):
looks ok.

morri-son (dismissed):

fabianburth (dismissed):

fabianburth (approved):

fabianburth (approved):

PR #379, opened 2023-06-13
Add Keyless Signing using Sigstore

### LINKED ISSUE #335: Integrate an option for keyless signing
Please evaluate and integrate a keyless signing process like cosign into ocm.
https://www.youtube.com/watch?v=d5vROySf1I8

**Why is this needed**:
Let's Encrypt has simplified the ease of consumption and life cyclemanagement for web certificates.
Similarly, sigstore.dev with the help of OpenSSF and the supply chain momentum will simplify the code signing process.

The value proposition is undeniable.

OCM should support and contribute to the emerging standard around Sigstore and offer a simple code signing experience.

phoban01:
## Description

This PR adds support for "keyless" signing using the [sigstore](https://docs.sigstore.dev) services Fulcio & Rekor.

Sigstore's Fulcio is used to generate short-lived signing certificates. In order to issue a signing certificate Fulcio requires an OIDC token. When the execution context is the users local machine it is possible to generate an OIDC token using sigstore's Dex instance. When the execution context is a GitHub Actions workflow the token will be taken from the workflow environment.

Once the component descriptor has been signed the signature is uploaded to Rekor and which returns a signed bundle that can be used to verify the validity of signature. We store the Rekor bundle in the signature value field which enables validation of signatures in offline or air-gapped environments.

Resolves #335 

## Usage
To add a signature using a short-lived signing certificate provided by Fulcio use the following command:

```shell
ocm sign component --signature test --algorithm sigstore --keyless ghcr.io/phoban01//phoban.io/test:v1.0.0
```

To validate the same signature we use the following:
```shell
ocm verify component --signature test --keyless ghcr.io/phoban01//phoban.io/test:v1.0.0
```


pkg/signing/handlers/sigstore/handler.go
```diff
@@ -0,0 +1,242 @@
...
+		FulcioURL:        "https://v1.fulcio.sigstore.dev",
+		OIDCIssuer:       "https://oauth2.sigstore.dev/auth",
+		OIDCClientID:     "sigstore",
+		SkipConfirmation: true,
+	}, signer)
+	if err != nil {
+		return nil, errors.Wrap(err, "new signer")
+	}
+
+	sig, err := fs.SignMessage(strings.NewReader(digest))
```
mandelsoft, line 81: I think here a digest from the digest is calculated. Is this a good idea?

phoban01: Yes that's correct. The input digest is a representation of the content so I don't think it's an issue.  

The other option here is to directly interact with the Fulcio server to request signing certificates and use those to sign the digest.

mandelsoft:
Hi @phoban01,

For the signer it is possible to circumvent the payload reader and to specify directly the digest. But your rekor upload handling
seems to require it.

I looked at the cosign project, how they do the rekor stuff. Here,  they are using `cosignv1.TLogUpload(ctx, r, sigBytes, checkSum, b)`to upload to rekor. And it does not seem to require the playload, but only a hash.Hash object which could be faked from a given digest, because here only the Sum call seems to be used. (see cosign/  internal/pkg/cosign/rekor/signer.go).

They are also using a `EntryToBundle` function to create a bundle from the rekor entry, which is then stored. Is there a reason why you did it differently (and don't create a bundle)?

phoban01:
I had a look at the cosignv1.TLogUpload, and I think we can borrow the approach the use to create the rekor entry but rest we don't need (https://github.com/sigstore/cosign/blob/9da30eacfca8c972dc188533b95688e0a23b7113/pkg/cosign/tlog.go#L233).

I didn't actually see the EntryToBundle function. As far I can see it simply remaps the rekor response payload into a cosign specific type. Also looking at the RekorBundle type definition it doesn't have any json tags so I get the feeling this is intended more for internal usage. Do you think we need it? I'm happy either way.

phoban01:
Updated to bypass hash calculation when signing in favor of using the existing digest. In tandem with this switching to  Rekor Hashed Rekord entry type. The entire process now uses the digests calculated by OCM directly 😄

morri-son (approved):
tested both signing and verification using an own component: works!

PR #380, opened 2023-06-14
add get inexpensive content version identity functionality

fabianburth:
The primary use case for this functionality is the caching of resources. Efficient caching requires a possibility to uniquely identify the content of a resource (so, an identity that changes if the content of the resource changes). Whether or not such an identity can be provided for a resource depends on the storage technology, or rather the access method (for oci artifacts within a oci registry, the manifest digest can be used, for blobs within an s3 storage a certain version attribute can be used, ...). This method provides a uniform interface to retrieve such an identity **if possible**. 

Fixes #

**Release Notes**:

```feature user
add inexpensive content version identity functionality to enable caching
```

### DIFF
--- a/pkg/contexts/ocm/accessmethods/plugin/plugin.go
+++ b/pkg/contexts/ocm/accessmethods/plugin/plugin.go
@@ -53,6 +53,46 @@ func (p *PluginHandler) AccessMethod(spec *AccessSpec, cv cpi.ComponentVersionAc
 	if mspec == nil {
 		return nil, errors.ErrNotFound(errors.KIND_ACCESSMETHOD, spec.GetType(), descriptor.KIND_PLUGIN, p.Name())
 	}
+
+	creddata, err := p.getCredentialData(spec, cv)
+	if err != nil {
+		return nil, err
+	}
+
+	info, err := p.Info(spec)
+	if err != nil {
+		return nil, err
+	}
+	return newMethod(p, spec, cv.GetContext(), info, creddata), nil
+}
+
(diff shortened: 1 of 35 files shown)


pkg/contexts/ocm/accessmethods/plugin/plugin.go
```diff
@@ -74,6 +76,47 @@ func (p *PluginHandler) AccessMethod(spec *AccessSpec, cv cpi.ComponentVersionAc
...
+	if creds != nil {
+		creddata, err = json.Marshal(creds)
+		if err != nil {
+			return "cannot marshal creds"
+		}
+	}
+	specdata, err := spec.GetRaw()
+	if err != nil {
+		return ""
+	}
```
mandelsoft, line 112: All this should be the same for the Reader implementation, Can we dedup this coding?

pkg/contexts/ocm/plugin/cache/doc.go
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Open Component Model contributors.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package cache implements the plugin cache that contains all loaded plugins. It keeps the metadata for the various
+// found plugins and provides a Go abstraction to call the appropriate commands.
```
mandelsoft, line 6: discovered instead of found

mandelsoft (changes requested):

mandelsoft (approved):

PR #389, opened 2023-06-22
add ocm image build to release workflow

### LINKED ISSUE #289: Add creation of Docker image containing the OCM client to the build / release process
- The build / release process should create a Docker image that contains the OCM client of that build / release. The image should be based on the alpine image (:latest or :3)
- The docker image should be published on ghcr.io with the name "com-client", tagged with the same version the OCM cli has, e.g. 0.2.0. It may make sense to have two tags, one that contains the base image provider and one without, e.g. ghcr.io/open-component-model/ocm-cli:0.2.0 and ghcr.io/open-component-model/ocm-cli:0.2.0-alpine
- There should some metadata about the image in the release notes
  - exact image name and SHA
  - what exact base image was used to build the image specifying the version and SHA the base image has at time of the build
  - link to the Dockerfile used
  - location of the ocm-client binary and maybe some example how to call it (docker exec <image> ocm-client ...
  - information like the link to the Dockerfile and more generic information like how to execute it, may be placed into a README that is placed in the ocm repository and then simply referenced in the release notes

In addition to ghcr.io we should check if it makes sense to upload the image to Dockerhub.

**Why is this needed**:
Many CI / CD pipelines execute actions using Docker images that contain the required tooling instead of pulling the tooling into the pipeline workspace. Having a Docker image that contains the OCM tooling, helps to support teams in using OCM.

A valid and tested Dockerfile with arguments for the alpine base image version and the ocm cli version would be:

```
ARG ALPINE_VERSION=latest
FROM alpine:${ALPINE_VERSION}
WORKDIR /tmp
ARG OCM_VERSION=v0.2.0
RUN wget https://github.com/open-component-model/ocm/releases/download/${OCM_VERSION}/ocm_Linux_x86_64.tar.gz
RUN tar xzf ocm_Linux_x86_64.tar.gz

ARG ALPINE_VERSION=latest
FROM alpine:${ALPINE_VERSION}
WORKDIR /root/
COPY --from=0 /tmp/ocm /usr/bin/ocm
```

jensh007:
add an OCI image with ocm CLI to release process

Fixes #289

**Release Notes**:

```feature user

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -40,7 +40,7 @@ jobs:
           v="v$(go run $GITHUB_WORKSPACE/pkg/version/generate --no-dev print-rc-version ${{ github.event.inputs.prerelease }})"
           if [ -n "${{ github.event.inputs.prerelease }}" ]; then
             echo "Candidate: $v"
-          else 
+          else
             echo "Candidate: $v (taken from source)"
           fi
         else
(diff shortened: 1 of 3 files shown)


morri-son (approved):
looks good. Let's try this out! :-)

jensh007:
Closing in favor of makefile based solution

PR #431, opened 2023-07-13
fix: length check in if statement for finalizers

Skarlso:

Fixes #

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/finalizer/finalizer.go
+++ b/pkg/finalizer/finalizer.go
@@ -180,7 +180,7 @@ func (f *Finalizer) Nested() *Finalizer {
 	f.lock.Lock()
 	defer f.lock.Unlock()
 
-	if f.nested == nil || f.nested.Length() > 0 {
+	if f.nested == nil || f.nested.Length() == 0 {
 		f.nested = &Finalizer{}
 	} else {
 		f.pending = append(f.pending[:f.index], f.pending[f.index+1:]...)
--- a/pkg/finalizer/finalizer.go
+++ b/pkg/finalizer/finalizer.go
@@ -180,7 +180,7 @@ func (f *Finalizer) Nested() *Finalizer {
 	f.lock.Lock()
 	defer f.lock.Unlock()
 
-	if f.nested == nil || f.nested.Length() == 0 {
+	if f.nested == nil || f.nested.Length() > 0 {
 		f.nested = &Finalizer{}
 	} else {
 		f.pending = append(f.pending[:f.index], f.pending[f.index+1:]...)
@@ -268,5 +268,8 @@ func (f *Finalizer) finalize() (err error) {
 		list.Add(f.pending[l-i-1]())
 	}
 	f.pending = nil
+	// just forget nested ones. They are finalized here, but are then invalid.
+	// Adding entries after the parent has been finalized is not supported, because there is no valid determinable order.
+	f.nested = nil
 	return list.Result()
 }
--- a/pkg/finalizer/finalizer_test.go
+++ b/pkg/finalizer/finalizer_test.go
@@ -268,6 +268,62 @@ var _ = Describe("finalizer", func() {
 			Expect(order).To(Equal(Order{"second", "next", "first", "closer", "nested"}))
 		})
 	})
+
+	Context("finalize ordering of parent and nested finalizers", func() {
+		It("handles empty", func() {
+			var f finalizer.Finalizer
+			f.Nested()
+			f.Nested()
+			f.Nested()
+		})
+		It("handles mixed nested and finalize", func() {
+			var f finalizer.Finalizer
+			var order Order
+
+			f.With(F("A", &order))
+			n := f.Nested()
+			n.With(F("B", &order))
+			f.Finalize()
+			f.With(F("C", &order))
+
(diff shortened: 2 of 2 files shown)


souleb (dismissed):

phoban01 (approved):

mandelsoft (approved):
Hi @Skarlso ,

unfortunately, it is not so easy. The code location you fixed was correct. The problem is caused
by a scenario where the nested finalizer is reused after the main finalizer has been finalized.
After this finalization the stored index is not valid anymore, because the pending list has been resetted.

There are two possible solutions:
- after the finalization is done, the stored nested one if just removed, because it has already been finalized. It should not be used afterwards, because it is invalidated after a parent finalize.
  I've added an appropriate commit with the tests
- keep the cached nested one at position 0 after finalization the parent. But this would result in a strange execution order, if it is still used by the original caller without re-requesting a new one. Therefore I think it is useful not to choose this kind of fix.

Skarlso:
Without the fix the tested scenario runs into and index out of bounds error. The check there shouldn't be checking if that slice has zero elements? It appears to be overriding things is they are NOT empty at the moment. Is that the intended behaviour?

mandelsoft:
The intended usage scenario is the following.

- creating a main finalizer
- calling defer for its Finalize method
- adding finalizations
- for loops: (also for nested loops (uses `Nested()` on the next outer finalizer instance)
  - create a nested finalizer for the actual parent
  - add loop-local finalizations
  - potentially add finalizations of objects, which should survive a loop step to an outer finalizer.
  - potentially return errors or final results (here the main defer will finalize everything still not yet finalized, over all loops)
  - finalize the loop local nested finalizer to release loop-local resources
  - end loop step
- continue on outer level potentially adding more finalizations to the outer finalizer
- end of method (defer will be executed and executes all pending finalizations)

So, when calling `Nested()` on a finalizer, this sub finalizer is intended to be finalized before the next call to `Nested()`.

Now to the implementation.

Whenever a nested finalizer is created its finalize method is added to the parent's finalization queue.
Basically for all calls a new Finalizer object could be created and put onto the queue. But this would mean, that loops with thousands of steps would produce a very long queue, although ever step should be finalized locally.

To avoid this, the nested finalizer is explicitly stored and the implementation tries to reuse it for each call.
Then two cases can happen:
- a previous finalizer is correctly finalized before the next call to `Nested()`. Then the stored object can be reused. But its location in the queue must be corrected, because in the meantime something could have been added to the parent, because it should survive the actual loop step. (this means !=nil but Length()==0), The actual entry in the queue (field index) if released and the (reused) object is append at the end
- a previous finalizer is NOT correctly  finalized before the next nested instance is requested. In this case, to ensure correct functionality, the implementation acts like no nested finalizer would have been used. It just keeps the old instance in the quere where it is and creates a new one, which is then added to the end of the queue. This is basically identical to the behaviour of always using regular `defer`s in the loop step coding. But the entry must be kept in queue to get finalized in the correct order. Your fix would break this contract.

So, the indexing problem can only occur, if a further nested one is requested after the parent has been finalized.
For example, you do something, then add loop handling, then you do something which requires finalization, then you continue using the finalizer for a new loop requesting nested ones.

In this case the finalization clears the queue and the stored index is not valid anymore, which causes the out-of-bounds exception. Therefore the simplest correct solution is just to forget the stored reusable nested finalizer object, so that for the next request and new one is created. This is exactly the missing line `f.nested = nil`.

Skarlso:
This is super error prone
> But its location in the queue must be corrected

If this is a linked list or a queue shouldn't you be doing pop or pull? I feel like a diffent data structure should be applied here to make this whole affair a lot less complex thereby removing the possibility that indexes need to be applied to find the right finalizer.

Skarlso:
>Whenever a nested finalizer is created its finalize method is added to the parent's finalization queue.
>Basically for all calls a new Finalizer object could be created and put onto the queue. But this would mean, that loops with thousands of steps would produce a very long queue, although ever step should be finalized locally.

I can't imagine a scenario where the finalizer queue would have THOUSANDS of steps in it. This feels like premature optimization here which warrants this complex flow. Do you have a use case or benchmark for such a situation?

Skarlso:
>In this case the finalization clears the queue and the stored index is not valid anymore, which causes the out-of-bounds exception. Therefore the simplest correct solution is just to forget the stored reusable nested finalizer object, so that for the next request and new one is created. This is exactly the missing line f.nested = nil.

I see what you are saying. That said, I believe that either numbers or valid use cases should defend / warrant this implementation here. A simple queue would most likely be enough here.

mandelsoft:
Hi @Skarlso 

I've already added the missing line. This is in method `finalize()`. The method `Nested()` is correct, the problem was in method `finalize()` where the pending list was resetted, but the nested object was not resetted, which led to the error when calling `Nested()` afterwards.

Forget the word queue. `pending` is just a list of pending finalizations. It is processed in the opposite order used to fill it.
The field index stored the index in this list used to place the nested finalizer object.

Skarlso:
Ah okay. Can you push that into this PR? Or are you creating a separate one?

Skarlso:
Ah sorry, for some reason, I didn't notice you already did. :) Thanks! 🙇

PR #441, opened 2023-07-16
Provide default implementation to support implementations of simple virtual/simulated OCM repositories,

mandelsoft:

This PR provides an OCM repository implementation based on two simple implementation interfaces, These interfaces 
can be implemented to provide virtual repositories based on some mapping of existing data structures.

It is intended to map OCM information provided in some context to be presented as complete OCM repository.
This will be used to map the inline OCM component descriptors optionally provided by a *Landscaper Installation*
to an OCM repository view. This enables to use those virtual environments exactly like a regular OCM repository context,
which avoids special cases for the regular installation processing coding.

An example implementation of these interfaces is based on some files in a virtual filesystem. It is used for the test cases, also.

This OCM repository wrapper based on those interface supports complete reference handling and readonly mode as well as
modifiable virtual environments.

Fixes #

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/repositories/virtual/repository.go
+++ b/pkg/contexts/ocm/repositories/virtual/repository.go
@@ -44,6 +44,9 @@ func (r *RepositoryImpl) Close() error {
 }
 
 func (r *RepositoryImpl) GetSpecification() cpi.RepositorySpec {
+	if p, ok := r.access.(RepositorySpecProvider); ok {
+		return p.GetSpecification()
+	}
 	return NewRepositorySpec(r.access)
 }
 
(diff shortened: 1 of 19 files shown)


pkg/contexts/ocm/repositories/virtual/repository.go
```diff
@@ -0,0 +1,69 @@
...
+}
+*/
+
+func (r *RepositoryImpl) Close() error {
+	return r.access.Close()
+}
+
+func (r *RepositoryImpl) GetSpecification() cpi.RepositorySpec {
+	return NewRepositorySpec(r.access)
+}
```
fabianburth, line 48: we should adjust this to enable proper serialisation of repository specs

fabianburth (changes requested):

pkg/contexts/ocm/repositories/virtual/repository.go
morri-son: @fabianburth @Uwe, is this ready for review/merge now?

mandelsoft: yes, I've fixed this

morri-son (approved):

fabianburth (approved):

mandelsoft:
rebased

fabianburth (approved):

PR #443, opened 2023-07-18
check for existing transfer target not matching the source

### LINKED ISSUE #387: ocm should fail when transfer to remote repo is skipped because it already exists.
**What happened**:
Try pushing with a component that already exists:
```
ocm transfer ctf gen/ctf ghcr.io/jensh007/ocm   
transferring component "ocm.software/ocmcli-image"...
  transferring version "ocm.software/ocmcli-image:0.3.0-rc.2"...
    version "ocm.software/ocmcli-image:0.3.0-rc.2" already present -> skip transport
...
echo $?                                         
0
```

exit code should be non zero

Note that the option `-f` is not set. Currently you have to look at stdout to see that it was skipped.

see above

I detected this when executing a github action that succeeds and you don’t have any clue that it actually did not do anything.
When someone really needs the current behavior I would prefer adding another option like `--ignore-failure-on-exist` or something. It probably comes from the fact that some components before might already have been pushed and the question arises then whether to continue with the remaining ones or abort. 

**Environment**:
ocm version 0.3.0-dev+4b053e6ddd15485b58e9808c3cf33a5034cff541

mandelsoft:

During a transfer a potentially already existing target version is checked to be equal to the source version before just omitting
the transport. If it is not equal an error is provided.

Fixes #387

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/compdesc/componentdescriptor.go
+++ b/pkg/contexts/ocm/compdesc/componentdescriptor.go
@@ -286,6 +286,7 @@ type ObjectMetaAccessor interface {
 // ElementMetaAccessor provides generic access an elements meta information.
 type ElementMetaAccessor interface {
 	GetMeta() *ElementMeta
+	IsEquivalent(ElementMetaAccessor) bool
 }
 
 // ElementAccessor provides generic access to list of elements.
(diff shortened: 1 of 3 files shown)


jensh007 (approved):
tested.

PR #449, opened 2023-07-19
fix component archive tests

fabianburth:
The test was wrong and overcomplicated.

Fixes #

**Release Notes**:

```feature user

```


mandelsoft (changes requested):
If you use the os filesystem you should not use .ACC_WRITABLE,

mandelsoft (approved):

Issue #457, opened 2023-07-24
Defer in the for loop in Transfer


Skarlso:
**What happened**:

https://github.com/open-component-model/ocm/blob/c3d77cc04d81e706fc31781ad4e679acec40837d/pkg/contexts/ocm/transfer/transfer.go#L132

```go
	for i, r := range src.GetResources() {
		var m ocm.AccessMethod
		a, err := r.Access()
		if err == nil {
			m, err = r.AccessMethod()
			if err == nil {
				defer m.Close()
				ok := a.IsLocal(src.GetContext())
	
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know**:

**Environment**:

PR #464, opened 2023-07-26
fix: transfer logic has defer in loop and nested error checks

Skarlso:

Fixes https://github.com/open-component-model/ocm/issues/457

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/transfer/transfer.go
+++ b/pkg/contexts/ocm/transfer/transfer.go
@@ -116,6 +116,21 @@ func transferVersion(printer common.Printer, log logging.Logger, state WalkingSt
 	return list.Add(comp.AddVersion(t)).Result()
 }
 
+type runContext struct {
+	err error
+}
+
+func run[T any](c *runContext, f func() (T, error)) T {
+	var result T
+	if c.err != nil {
+		return result
+	}
+
+	result, c.err = f()
+
+	return result
+}
+
 func CopyVersion(printer common.Printer, log logging.Logger, hist common.History, src ocm.ComponentVersionAccess, t ocm.ComponentVersionAccess, handler transferhandler.TransferHandler) error {
 	if handler == nil {
 		handler = standard.NewDefaultHandler(nil)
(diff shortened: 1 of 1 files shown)


pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -195,10 +236,11 @@ func printArtifactInfo(printer common.Printer, log logging.Logger, kind string,
 		} else {
 			printer.Printf("...%s %d...\n", kind, index)
 		}
-	}
-	if hint != "" {
-		log.Debug(fmt.Sprintf("handle %s", kind), kind, fmt.Sprintf("%d(%s)", index, hint))
 	} else {
-		log.Debug(fmt.Sprintf("handle %s", kind), kind, fmt.Sprintf("%d", index))
+		if hint != "" {
```
Skarlso, line 240: I think this was double logging. Was that intentional? If printer was defined it would still log again on log.Debug.

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -124,65 +139,93 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
...
-				if !ok {
-					if !none.IsNone(a.GetKind()) {
-						ok, err = handler.TransferResource(src, a, r)
-						if !ok {
-							log.Info("transport omitted", "resource", r.Meta().Name, "index", i, "access", a.GetType())
-						}
+		c := &runContext{}
+
+		// gather the access and the access method
+		a := run(c, r.Access)
```
souleb, line 145: it is safe to use the same `runContext` and overwrite the error? maybe join them instead?

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -124,65 +139,93 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
...
+		m := run(c, r.AccessMethod)
+
+		if c.err == nil {
```
souleb, line 148: why not start by failing early like:

```go
		if c.err != nil {
			if m != nil {
				m.Close()
			}

			if !errors.IsErrUnknownKind(c.err, errors.KIND_ACCESSMETHOD) {
				return errors.Wrapf(c.err, "%s: transferring resource %d", hist, i)
			}

			printer.Printf("WARN: %s: transferring resource %d: %s (enforce transport by reference)\n", hist, i, c.err)
		} else {
			...
		}
```

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -124,65 +139,93 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
...
+			// so far so good, let's perform the next set of actions
+			ok := a.IsLocal(src.GetContext())
+			if !ok {
+				if !none.IsNone(a.GetKind()) {
```
souleb, line 152: if find having a pkg named `none` troubling.

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -124,65 +139,93 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
...
-						}
+		c := &runContext{}
+
+		// gather the access and the access method
+		a := run(c, r.Access)
+		m := run(c, r.AccessMethod)
+
+		if c.err == nil {
+			// so far so good, let's perform the next set of actions
+			ok := a.IsLocal(src.GetContext())
```
souleb, line 150: can we have a variable that reflect what we expect, like `sourceReady`?

pkg/contexts/ocm/transfer/transfer.go
Skarlso: The purpose is to not do anything if there was an error previously.

pkg/contexts/ocm/transfer/transfer.go
mandelsoft: Hmm, logging is logging and printing is printing. The log messages, if enabled, are sent to a log output. The printer is feedback for the user  (stdout for CLI, or explicit printer setting for a lib user).

mandelsoft (changes requested):
Please have a look at #457.

This should be the simple way to handle defers in loops.

pkg/contexts/ocm/transfer/transfer.go
Skarlso: Okay. I'll leave that as is then. :)

PR #465, opened 2023-07-26
fix: transfer logic has defer in loop

### LINKED ISSUE #457: Defer in the for loop in Transfer

mandelsoft:

Fixes #457

**Release Notes**:

```feature user

```

### DIFF
--- a/hack/install-requirements.sh
+++ b/hack/install-requirements.sh
@@ -9,7 +9,7 @@ PROJECT_ROOT="${CURRENT_DIR}"/..
 
 echo "> Install Go packages/binaries"
 
-curl -sfL "https://install.goreleaser.com/github.com/golangci/golangci-lint.sh" | sh -s -- -b $(go env GOPATH)/bin v1.32.2
+curl -sfL "https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh" | sh -s -- -b $(go env GOPATH)/bin  v1.49.0
 
 go install github.com/go-bindata/go-bindata/v3/go-bindata@v3.1.3
 go install golang.org/x/tools/cmd/goimports@latest
--- a/pkg/contexts/ocm/transfer/transfer.go
+++ b/pkg/contexts/ocm/transfer/transfer.go
@@ -19,7 +19,6 @@ import (
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/transfer/transferhandler"
 	"github.com/open-component-model/ocm/pkg/contexts/ocm/transfer/transferhandler/standard"
 	"github.com/open-component-model/ocm/pkg/errors"
-	"github.com/open-component-model/ocm/pkg/finalizer"
 	"github.com/open-component-model/ocm/pkg/runtime"
 )
 
@@ -118,9 +117,6 @@ func transferVersion(printer common.Printer, log logging.Logger, state WalkingSt
 }
 
 func CopyVersion(printer common.Printer, log logging.Logger, hist common.History, src ocm.ComponentVersionAccess, t ocm.ComponentVersionAccess, handler transferhandler.TransferHandler) (rerr error) {
-	var finalize finalizer.Finalizer
-	defer finalize.FinalizeWithErrorPropagation(&rerr)
-
 	if handler == nil {
 		handler = standard.NewDefaultHandler(nil)
 	}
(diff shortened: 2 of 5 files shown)


pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -116,7 +117,10 @@ func transferVersion(printer common.Printer, log logging.Logger, state WalkingSt
 	return list.Add(comp.AddVersion(t)).Result()
 }
 
-func CopyVersion(printer common.Printer, log logging.Logger, hist common.History, src ocm.ComponentVersionAccess, t ocm.ComponentVersionAccess, handler transferhandler.TransferHandler) error {
+func CopyVersion(printer common.Printer, log logging.Logger, hist common.History, src ocm.ComponentVersionAccess, t ocm.ComponentVersionAccess, handler transferhandler.TransferHandler) (rerr error) {
+	var finalize finalizer.Finalizer
```
Skarlso, line 121: I don't understand this complexity. If you want it simpler then why isn't it enough to just plain call `m.Closer()`?

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -125,11 +129,13 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
 	log.Info("  transferring resources")
 	for i, r := range src.GetResources() {
 		var m ocm.AccessMethod
+		loop := finalize.Nested()
+
 		a, err := r.Access()
 		if err == nil {
 			m, err = r.AccessMethod()
 			if err == nil {
-				defer m.Close()
```
Skarlso, line 132: Another easier option to fix this would be to simply extract this part of the function into another function in which you can defer close the AccessMethod once done.

Skarlso:
This is a more readable approach https://github.com/open-component-model/ocm/pull/466

pkg/contexts/ocm/transfer/transfer.go
mandelsoft: You are right, in this case it could be solved easier, because there is only one place where close must be called.

But the example should just show, how it can be solved  (especially for more complex ones), including the correct forwarding of error codes. just a `defer x.Close()` is basically wrong wherever it is used, because it omits the return code, nevertheless everyone uses this pattern. Go does not really has a good formal way to handle this general problem, (like the loop/defer problem).

By the way, checking the the error of the close uncovered another problem in the code, Colose is called twice.

hack/install-requirements.sh
```diff
@@ -9,7 +9,7 @@ PROJECT_ROOT="${CURRENT_DIR}"/..
 
 echo "> Install Go packages/binaries"
 
-curl -sfL "https://install.goreleaser.com/github.com/golangci/golangci-lint.sh" | sh -s -- -b $(go env GOPATH)/bin v1.32.2
+curl -sfL "https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh" | sh -s -- -b $(go env GOPATH)/bin  v1.49.0
```
Skarlso, line 12: What prompted this change? It doesn't seem to be relevant to this PR.

pkg/contexts/ocm/transfer/transfer.go
```diff
@@ -125,25 +125,27 @@ func CopyVersion(printer common.Printer, log logging.Logger, hist common.History
...
-					printArtifactInfo(printer, log, "resource", i, hint)
-					err = handler.HandleTransferResource(r, m, hint, t)
-				}
+			}
+			if ok {
+				hint := ocmcpi.ArtifactNameHint(a, src)
+				printArtifactInfo(printer, log, "resource", i, hint)
+				err = errors.Join(err, handler.HandleTransferResource(r, m, hint, t))
+			} else {
+				err = errors.Join(err, m.Close())
```
Skarlso, line 148: `m` is not closed in all cases, only here. So if it is `ok` it won't be closed. Why not put this at the end? Since we only ever return at the end.

Skarlso (changes requested):

hack/install-requirements.sh
mandelsoft: yes, but the one who changed the linter settings never updated this. And I had to update the linter to get the same messages, which were provided in the lintercheck of this PR.

Skarlso: Gotcha 👍

pkg/contexts/ocm/transfer/transfer.go
mandelsoft: In the ok==true case the handler is called. And here the contract says, that the given access (m) is closed.
That was the reason, why m was closed twice. 

So, for ok==true the responsibility to close the access is passed to another entity (the handler), and
for the other case it has to be closed explicitly.

I agree, that is not really nice.

Skarlso: Ah, you mean this bit?
```go
func (b *blobAccess[T]) Close() error {
	b.lock.Lock()
	defer b.lock.Unlock()
	if !b.closed.Load() {
		tmp := b.access
		b.closed.Store(true)
		return tmp.Close()
	}
	return ErrClosed
}
```

I see. Well, this isn't ideal, indeed.

Skarlso (approved):

pkg/contexts/ocm/transfer/transfer.go
mandelsoft: I meant this implementation-induced take-over of the ownership of the access. I've changed this, now the caller (the transfor code) opens the access and has to close it again, so, the `Close` call only in the `else` part of an `if` can now be avoided. The access has always to be closed. The handler does not implicitly close it, now.

PR #466, opened 2023-07-27
fix: defer in loop statement during transport

Skarlso:

Fixes https://github.com/open-component-model/ocm/issues/457

**Release Notes**:

```feature user

```


mandelsoft (commented):
I think, this does not help. The interface of the extracted method is much too broad, because it uses lots of variables of the initial method.

Skarlso:
Well, the calling function also has a lot of parameters, so it pretty much blends in with the rest of the code.

That said, the parameters can be put into a struct and assigned like that.

PR #467, opened 2023-07-27
adjust ca handling for helm repositories

fabianburth:
To implement a proper handling of custom certificate authorities for interaction with helm repositories.

Fixes #

**Release Notes**:

```feature user

```


pkg/helm/downloader.go
```diff
@@ -139,8 +138,8 @@ func (d *chartDownloader) complete(ctx oci.ContextProvider, ref, repourl string)
 		return err
 	}
 
-	if len(d.cacert) != 0 {
-		err = d.writeFile("cacert", config, &entry.CAFile, d.cacert, "CA file")
+	if len(creds[identity.ATTR_CERTIFICATE_AUTHORITY]) != 0 {
+		err = d.writeFile("ca", config, &entry.CAFile, []byte(creds[identity.ATTR_CERTIFICATE_AUTHORITY]), "CA file")
```
mandelsoft, line 142: may be we should keep the filename `cacert`

mandelsoft (changes requested):

pkg/helm/chartaccess.go
```diff
@@ -50,17 +50,23 @@ type chartAccess struct {
 
 var _ ChartAccess = (*chartAccess)(nil)
 
+func newChartAccess(path string, fss ...vfs.FileSystem) (*chartAccess, error) {
```
Skarlso, line 53: What is the purpose of this extracted function? It isn't called anywhere else other than the temp access below which already does the thing this does. :) Is there a plan to call this from something else?

Skarlso: Neat. Thanks.

fabianburth: Initially, we didn't realise we had access to the file system of the chart access from the downloader. Therefore, we thought we needed an option to pass in an arbitrary file system. This is pretty much a remnant of that original thought. As it is currently not necessary, I removed it for now.

mandelsoft (approved):

PR #470, opened 2023-07-29
Provide wrapper function for often used OCM code sequences to simplify usage

### LINKED ISSUE #454: Better discoverability for public API usage
**User Story**: As a user for the Library, I would like to be able to intuitively figure out what functions to call to perform what opertaions.

**Description**:

Consider the following Signing example:

```go
		opts := signing.NewOptions(
			signing.Sign(ocmsigning.DefaultHandlerRegistry().GetSigner(rsa.Algorithm), component.Sign.Name),
			signing.Resolver(resolver),
			signing.PrivateKey(component.Sign.Name, component.Sign.PrivKey),
			signing.Update(), signing.VerifyDigests(),
		)

		if err := opts.Complete(signingattr.Get(c)); err != nil {
			return fmt.Errorf("failed to complete signing: %w", err)
		}

		if _, err := signing.Apply(nil, nil, component, opts); err != nil {
			return fmt.Errorf("failed to apply signing: %w", err)
		}
```

**Problems**:

- It's not clear that I have the call `signing.Apply` with very specific options to sign a component.
- Verification is the same but with different options. It's still apply.
- The options are hard if even to understand easily, until the user of the library has a really deep understanding of the signing process
- Why is `signing.Update()` needed? What does it do? Why does the user need to know about it? ( these are rhetorical questions, it shouldn't be there )

**Solution**:

A user-friendlier approach would be to either call sign on the component version like this:

```go
	cv, err := repo.LookupComponentVersion(name, version)
	if err != nil {
		return nil, fmt.Errorf("failed to look up component Version: %w", err)
	}
       if err := cv.Sign(privateKey); err != nil {}
```

Since we already have the component version, it contains all the context it needs to sign itself. This would be intuitive since I'm signing that specific version. Functional options could still be applied here optionally, like specifying the algorithm, or further details. The function should use sensible defaults internally if nothing is defined.

Verification could also be this simple, using something like `cv.Verify(publicKey)`.

Alternatively, the Sign method could accept the component version if we don't want to change the interface.

**Epic link**:
- #390

**Acceptance criteria**:

- [ ] easy to discover functionality for specific user scenarios

**Related issues**:

mandelsoft:

Fixes #454

**Release Notes**:

```feature user

```


morri-son:
@Skarlso , before the refactoring this would already help. Can you please review?

Skarlso:
@Skarlso Ping.

pkg/contexts/ocm/transfer/transferhandler/standard/options.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/open-component-model/ocm/pkg/utils"
 )
 
+func init() {
+	transferhandler.RegisterHandler(1000, &TransferOptionsCreator{})
```
Skarlso, line 18: How come this is suddenly needed here?

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
```diff
@@ -11,12 +11,65 @@ import (
...
+const KIND_TRANSFEROPTION = "transfer option"
 
+// TransferOptions is the general type for an option bag
+// holding options for a transfer handler.
+// Different transfer handler implementations may use differ
+// concrete option sets.
+// Therefore, every option set uses its own options type
+// The options setters/getters MUST be interfaces, which may be implemented
+// by different option set types to enable the option implementations
+// to work with different options sets.
```
Skarlso, line 23: This makes very little sense for me as a user of the library.

What is an option bag? Why is it a bag? Why would I care about all of these? Why is this introduced right now?

Is there a simpler way to do all of this?

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
```diff
@@ -11,12 +11,65 @@ import (
...
+type TransferOptions interface {
+	NewTransferHandler() (TransferHandler, error)
+}
+
+type TransferOptionsCreator interface {
+	NewOptions() TransferOptions
+}
+
+// TransferOption is an option used to configure a transfer handler.
```
Skarlso, line 32: Please try and simplify all of this. The bag synonym here does not help the explanation or the usage or the importance of these options that can apply themselves to option bags that then do... what?

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
```diff
@@ -11,12 +11,65 @@ import (
...
+func (o SpecilizedOptionsCreator[P, T]) NewOptions() TransferOptions {
+	var opts T
+	return P(&opts)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+// TransferHandler controls the transfer of component versions.
+// It can be used to control the value transport of sources and resources
+// on artifact level and the way how nested component version are transported.
```
Skarlso, line 68: >on artifact level

What is on artifact level? What's an artifact in this context?

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
```diff
@@ -11,12 +11,65 @@ import (
...
+// There are two implementations delivered as part of the OCM library:
+//   - package transferhandler.standard: able to select recursive transfer
+//     general value artifact transport.
+//   - package transferhandler.spiff: controls transfer using a spiff script.
```
Skarlso, line 72: What's a spiff script? A short explanation here would help.

pkg/contexts/ocm/utils/resource.go
```diff
@@ -16,3 +20,51 @@ func GetResourceData(acc ocm.ResourceAccess) ([]byte, error) {
 	defer m.Close()
 	return m.Get()
 }
+
```
Skarlso, line 23: Hmmm. This seems to be an unrelated change? And these don't appear to be used. What is the purpose of these?

pkg/contexts/ocm/transfer/autohandler.go
```diff
@@ -0,0 +1,111 @@
...
+func NewTransferHandler(list ...transferhandler.TransferOption) (h transferhandler.TransferHandler, ferr error) {
+	var opts transferhandler.TransferOptions
+
+	created := -1
+outer:
+	for {
+		ferr = nil
+		for i := 0; i < len(list); i++ {
+			o := list[i]
+			if o == nil {
```
Skarlso, line 53: This is interesting. How would this be possible? Why would someone pass in a nil option? Or would the option somehow evaluate to nil? This seems to be a bug then.

pkg/contexts/ocm/transfer/autohandler.go
```diff
@@ -0,0 +1,111 @@
...
+				continue
+			}
+			if opts == nil {
+				created = i
+				opts = o.NewOptions()
+			}
+			if err := o.ApplyTransferOption(opts); err != nil {
+				if errors.IsErrNotSupportedKind(err, transferhandler.KIND_TRANSFEROPTION) {
+					if i <= created {
+						// give a second chance to later options
```
Skarlso, line 63: Why? What would change between the handlers and the options that would make it pass on a second try? Why not a third or a forth? These seem like an arbitrary thing and doesn't help in understand what's going on here.

pkg/contexts/ocm/transfer/autohandler.go
```diff
@@ -0,0 +1,111 @@
...
+// (an option always belongs to a dedicated option set of a dedicated handler
+// but may be supported by other option sets, also) the options sets of all
+// registered handler types with checked by trying the most significant
+// handlers, first, to find the best matching handler for the given option set.
+func NewTransferHandler(list ...transferhandler.TransferOption) (h transferhandler.TransferHandler, ferr error) {
+	var opts transferhandler.TransferOptions
+
+	created := -1
+outer:
+	for {
```
Skarlso, line 49: Please try to simplify this three times nested for loop structure by extracting some of the inner components into descriptive functions. Incidentally, that would allow you to skip the labeling of the functions.

Also, the `created` index holder to identify state of the options is hard to follow. Try using a different data structue that would allow for easier reading and following the selection logic of the options.

I feel like the final part of the for loop could be a plain fallback logic on the registered options if none of the given options applied. I'm not sure what makes a second run possible and what would change that warrants only a single retry instead of plainly falling back.

pkg/contexts/ocm/transfer/frontend.go
```diff
@@ -0,0 +1,27 @@
...
+
+// TransferComponentVersionWithHandler uses the specified transfer handler to control
+// the transfer process.
+func TransferComponentVersionWithHandler(pr common.Printer, cv ocm.ComponentVersionAccess, tgt ocm.Repository, handler transferhandler.TransferHandler) error {
+	return TransferVersion(pr, nil, cv, tgt, handler)
+}
+
+// TransferComponentVersion uses the transfer handler based on the given options to control
+// the transfer process. The default handler is the standard handler.
+func TransferComponentVersion(printer common.Printer, cv ocm.ComponentVersionAccess, tgt ocm.Repository, optlist ...transferhandler.TransferOption) error {
```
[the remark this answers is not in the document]
mandelsoft: The repository is the intended target repository for the transfer, how could this be inferred? FRom which other given information?

The printer can be set to nil. It is intended by catch processing information, typically sent to a user.
r could this add this as additional option for the option list. I'll try to do this.

The problem hereby is, that the options so far are only used for transfer handler options. But this would a free floating options, which has to be considered by the convenience methods, only

pkg/contexts/ocm/transfer/init.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Open Component Model contributors.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package transfer
+
+import (
+	_ "github.com/open-component-model/ocm/pkg/contexts/ocm/transfer/transferhandler/config"
```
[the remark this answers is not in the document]
mandelsoft: The reason for those _ imports are always the same in go: just initialize the package, even if it not used to execute the init methods, which typically register implementation variants.

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 
 require (
 	github.com/Masterminds/semver/v3 v3.2.1
+	github.com/aws/aws-sdk-go v1.44.288
```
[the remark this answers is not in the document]
mandelsoft: Hmm, this is a problem. I've local test coding not checked in. This one seems to use this package. I don't know how I can differentiate between those use cases in Go

pkg/contexts/ocm/utils/resource.go
mandelsoft: I changed the contract to make it easier, but then I have to do an explicit  close, here.,

pkg/contexts/ocm/transfer/autohandler.go
mandelsoft: That is a typical case of defensive programming. Depending on the composition of the options by the caller it should be possible to handle a nil option, also. This will avoid unnecessary Panics and makes the live easier for the caller.

For example, he composes the options by a set of optional command line options, which will be provided in this library typically by option objects, which can be obtained from an option list of a command with <code>optionpackage.From(list)</code>.
This will return nil, if the option is not configured for the actually handled command.

pkg/contexts/ocm/transfer/autohandler.go
mandelsoft: Depending on the order of the given option list, it might happen that later options relate to a more specialized list.
Therefore every later option, if not matching the actual option target gets a (second compared to the first failing) chance to provide a better option target able to consume this failing option setter.

This means every option has
- a first chance, to configure the already known option target
- and a second chance to provide an own target object, it is not applicable to the actual target

pkg/contexts/ocm/transfer/autohandler.go
mandelsoft: The problem is the intrinsic complexity. If the logic of complete sub-loops is split into separate functions, the contract will be very broad, because it has to modify outer state. Additionally the final return is in the inner loops, which makes it difficult to handle this cross function-calls. The result will be much harder to understand than the actual solution.

mandelsoft:
Hi @Skarlso, I've changed the api to get rid of the required Printer parameter. This is now a additional options.
For the transfer, the option handling has been rearranged to support handler options and processing options. The interfaces have been adapted accordingly. The Option interface is now more general and an optional interface is implemented by handler options. to indicated its usage for handler options.

The frontend.go files have been renamed to convenience.go.

Skarlso:
Hello. :) Did you see my other comments? Sometimes Github hides them like this ->

Screenshot 2023-08-29 at 7 36 47

pkg/contexts/ocm/transfer/frontend.go
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Open Component Model contributors.
```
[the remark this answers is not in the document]
mandelsoft: frontend for *user-facing*, but may be convenience is better, I've changed it accordingly.

pkg/contexts/ocm/transfer/frontend.go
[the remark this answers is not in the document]
mandelsoft: I've moved this argument to an additional possibility for an option, The default now is /dev/null.

pkg/contexts/ocm/transfer/init.go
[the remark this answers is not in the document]
mandelsoft: In this case it registers the config object type usable to configure the handler.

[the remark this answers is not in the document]
Skarlso: I know what the purpose is in Go. :) I was thinking what is the purposes here exactly? In Go, normally, if you have a _blank_ import, you add a comment that explains the need for it. Like the SQL package in go that imports driver implementation.

go.mod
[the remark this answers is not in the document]
mandelsoft: I managed to remove it even for my local `go mod tidy`-

pkg/contexts/ocm/transfer/autohandler.go
Skarlso: This seems like a convoluted way to provide fall-back. But okay.

pkg/contexts/ocm/transfer/autohandler.go
Skarlso: 👍

pkg/contexts/ocm/signing/frontend.go
```diff
@@ -0,0 +1,69 @@
...
+		opts.Signer = signing.DefaultHandlerRegistry().GetSigner(rsa.Algorithm)
+	}
+	err := opts.Complete(nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "inconsistent options for signing")
+	}
+	return Apply(printer, nil, cv, &opts)
+}
+
+func VerifyComponentVersion(printer common.Printer, cv ocm.ComponentVersionAccess, name string, pubkey interface{}, optlist ...Option) (*metav1.DigestSpec, error) {
```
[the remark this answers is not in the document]
mandelsoft: I've moved the printer argument to an additional possible options.

My intention was to offer a function with parameter usually required for signing/verification,

To put everything into the options just to achieve less parameters was just the approach of the original Apply functions,
which is the _swiss knife_ for signing.

But I remove the keys now, again. Nevertless, typically, a key is required for signing/verification.
If you don't specify a key it must be configured in the context and this is even more implicit than the specification per option.,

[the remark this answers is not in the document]
Skarlso: Gotcha. Thanks.

pkg/contexts/ocm/signing/options.go
```diff
@@ -93,6 +93,15 @@ func Sign(h signing.Signer, name string) Option {
 	return &signer{h, name}
 }
 
+func Signer(h signing.Signer) Option {
```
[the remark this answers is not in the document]
mandelsoft: I added docu for all signing options.

pkg/contexts/ocm/signing/options.go
```diff
@@ -258,10 +267,12 @@ func PrivateKey(name string, key interface{}) Option {
 }
 
 func (o *privkey) ApplySigningOption(opts *Options) {
-	if opts.Keys == nil {
-		opts.Keys = signing.NewKeyRegistry()
+	if o.key != nil {
```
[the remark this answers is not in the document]
mandelsoft: I'm not convinced, that this always improves readability, but here it's ok.

pkg/contexts/ocm/transfer/transferhandler/registry.go
```diff
@@ -0,0 +1,62 @@
...
+
+var optionsets = &registry{}
+
+// RegisterHandler registers handler specific option set type
+// for option based handler detection.
+// Self-contained handlers should use priority 1000.
+// Handlers incorporating extending option sets from other handlers
+// should use priority 100.
+// Handlers supporting the union of incompatible option sets
+// should use lower priority according to the order they should be checked.
```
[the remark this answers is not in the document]
mandelsoft: I'll add a package doc explaining the purpose of transfer handlers and improve the doc above.

pkg/contexts/ocm/transfer/transferhandler/standard/options.go
mandelsoft: This is the fallback transfer option set handler used in a second stage to map a set of given options to an appropriate
transfer handler. 

THis mechanism is only used for standard handlers used to operation flavors only taking transfer options and no transfer handler.  The the set of given options is used to determine the best matching handler implementation.
This simplifies the convenience operations, they just take options and the caller does not need to deal with transferhandler selection and conguration.

 If a custom handler is used, another operation flavor directly taking a transferhandler should be used.

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
mandelsoft: I'll add an explanation, but basically a typical user never has to deal with this type. This in only relevant for implementing standard handlers with their own set of options.

To clearly separate those use cases, it would be required to introduce multiple packages and because of the arising dep cycles, again an internal package would be required, which would complicate the package structure even more. Because of this I've put everything, the user relevant parts and the implementation specific parts into a single package,

mandelsoft: I'll support this separation of concern by making the transfer package self-contained for the typical use case.
So, it will provide all the types required to use it. There is no need anymore to deal with the transferhandler package in the import list.

If the user wants to explicitly use a standard handler he can use the appropriate handler package.
Only if the user wants to provide an custom handler  to implement custom logic, he has to deal with the transferhandler package.

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
mandelsoft: See above, the transfer handler package is now only for implementors of transfer handlers.

The option handling is explained in detail and is only relevant, if the user wants to provide an own (may be overlapping) set of options.

pkg/contexts/ocm/transfer/transferhandler/transferhandler.go
mandelsoft: Any decision about the process can be based of attributes of the described artifact (common term for resource and source, see ocm-spec).

mandelsoft:
Hi @Skarlso, I hope I've covered all issues now. The latest commit contains the described changes.

Skarlso (approved):
LGTM 👍

Issue #469, opened 2023-07-28
Transporting CV graphs with CVs shared with separately transported graphs causes problems


mandelsoft:
**What happened**:

If there are two aggregated component versions sharing a common CV, transportation is hardly possible for signed components.

- even for unsigned CVs the transport fails for the transport of the second CV, because the shared one is already there.
  The check (basically a tweaked deep equal) for valid existing components fails, because the access methods might be touched.
  So far, it is not possible to detect whether a component is really the same without reading and comparing the blobs of both variants.
  The only solution so far is the option -f, which overwrites the existing CVs in the target
- just overwriting is no solution because:
  - it should be able to fail if the shared CV contains different technical resources
  - digests and signature would be lost if the second transport contains no or different digests
  - what should happen is both aggregated CVs are digested with different algorithms.

**What you expected to happen**:

It must be possible to transport partly shared CV graphs, which are independently signed in a safe way, guaranteeing that all possible
transportation sequences finally provide the same result.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know**:

**Environment**:

mandelsoft:
Basically, we urgently have to rework the creation component versions.

First: The only way to attack this problem is to provide the necessary resource digests as early as possible: during the CV creation.

This means:
- the digest algorithm is determined already when creating a CV
   -> no different algorithms possible for different signing sequences for different aggregation CVs
   -> during the transport it is possible to check the identity of resources, just by comparing the digest.
   -> creating with access instead of input will be more expensive, because all blobs must be read
   -> creating resources by access will only be possible if the appropriate access type is known to the actual version
        of the toolset
- add check to equality to transport if CV is already present at target based on the now available digests.
  - fail if content differs and accept is content is identical
  - problem: transport of signatures

Second: During transport the signatures must be merged. Even if the CV describes the same content, the signatures in the
  target must potentially be adapted to reflect both, the target and the source signature.

Problem: how to handle resigning with the same signature identity?
  Transporting should update the signature -> we need a signing timestamp to identify the latest one. Is this really secure?

mandelsoft:
The following scenario must be handled correctly, also:
- an already transported CV is locally signed in the target environment
- an update with (non-signature relevant changes, for example routing slip is imported)...
- ...with potentially additional signatures.

Here, also the early digesting and the proposed merge operation could help to support this scenario.

PR #478, opened 2023-08-04
fix provider normalization

### LINKED ISSUE #477: CD normalization does not handle provider labels correctly
**What happened**:

In V3 (and the internal CD representation) a complex provider object is possible, with name and labels.

These labels are not correctly normalized because the signature property is not obeyed to select the signature-relevant labels, instead all labels are assumed to be signature-relevant.

In CD V2 complex provider information is transformed to a json string value of the provider (string) field. Here, the label normalization is not handled, correctly. Just the complete provider string is used.
 

**Environment**:

mandelsoft:

The provider labels in the component descriptor are not normalized correctly.
For details, see the issue #477 

This fix now assures, that the provider labels are normalized correctly in all combinations (norm algo/serialization format).
- norm/v1 CD v2
- norm,/v1 CD v3
- norm/v2

Relates to #469.

Fixes #477

**Release Notes**:

```feature user

```


morri-son (approved):

mandelsoft:
Wxample:  https://github.com/open-component-model/ocm/blob/a45c3c0527da39f9d1707b08d4fc8ca521176ea5/pkg/contexts/ocm/compdesc/norm_test.go#L210-L246

PR #483, opened 2023-08-04
Transfer

mandelsoft:

Adding resources to a Component Version now already adds digests by default.
This enables delta transports of volatile data in the Component Descriptor.
This volatile data can now be merged: Local (volatile)  changes in the target repository
of a transfer are merges with the new inbound version.

This is mainly used to merge non-signature-relevant label values. Therefore,
value merge algorithms are introduced, which can be declared in the component descriptor
or pre configured in the code. The standard library provides a set of basic algorithms
with appropriate configuration options, which can be extended by plugins.

This fatur can, for example , be used for implementing routing slips on top of the
Open Component Model in form of dedicated labels. The delta transport feature is then used
to re-import/transport new routing slip entries from the software provider.

To be compatible, the handling of resources without a digest is supported, also,but with 
limited functionality.

A possible (not yet implemented) extension would be to introduce a new transfer option
to enable on-the-fly calculation of missing digests.
This will always enable to skip the value transfer of resource versions already available at the target,
if the overwrite option is chosen.

Fixes #

Relates to #469 

**Release Notes**:

```feature user

```


mandelsoft:
Hi @fabianburth , changes from the past days pulled down an rebased to main.

fabianburth (approved):

PR #511, opened 2023-09-20
Routingslips as dedicated Label

mandelsoft:

This PR introduces a basic support for routing slips. Routing slips offer the possibility
to add a sequence of  information to a components version. For example, information about executed process steps
and their results. Such a sequence will be independently (from component version signatures) signined. A component version
may carry multiple routing slips from different providers.

They are implemented as non-signature-relevant label using the new delta transfer mechanism supporting
to re-transfer new entries to a target repository.

Every routing slip entry has a formal type (like labels). There is a standard type (comment) but any
arbitrary content may be added, also, with the new `routingslip` command section for the CLI.
It supports command line options to compose the content of an entry (like for access methods in the `add` commands)
New formal types with (new) command line options can be added by plugins. But even without using such formally
known types, arbitrary types with the data can be used (by providing appropriate yaml/json content.

Additionally some new features are added:
- private key encryption. Key files can be encrypted by a chain of other keys.
- OCI artifact download is more flexible now (for example downloading a dedicated layer, only)

Several fixes:
- The composition option handling is fixed for various special scenarios
  - The version support for composition options now works
  - Error message for given composition options not required/supported by the selected type is fixed.
- The plugin frame implementation is now able to handle complex composition options.

Library features:
- signing/hash attributes are now bound to the OCM context instead of always using global registries.

Fixes #

**Release Notes**:

```feature user

```a


mandelsoft:
reviews together with @fabianburth. Review results incorporated with last commit.

fabianburth (approved):

PR #516, opened 2023-09-21
consider transport modes for transport flow decisions

### LINKED ISSUE #513: Merge of PR #483 breaks integration tests
Since merging the PR integration tests are failing:
See e.g. [Run log](https://github.com/open-component-model/ocm-integrationtest/actions/runs/6256334280/job/16986997486)

I can reproduce the error also when running the tests locally. First analysis:

- It seems that he behavior of `ocm push -f --copy-resources` has changed

Input is like:
```
    components:
    ...
        resources:
          - name: image
            type: ociImage
            version: "1.10"
            access:
                type: ociArtifact
                imageReference: gcr.io/google_containers/echoserver:1.10
...
``` 
After pushing with the mentioned parameters the remote comp-descr in the oci registry contains in `access/imageReference`a URL. Now the host name and path is different after PR-merge.
before: `'<host>:<port>/inttest/google_containers/echoserver:1.10'` 
now: `'gcr.io/google_containers/echoserver:1.10'`

mandelsoft:

To decide, which resources have to be transported the digest is not sufficient to handle re-transports
of exactly the same content with different transport modes (value vs reference).

This Pr checks, whether access methods would have to be changed according to the
the actual state in the target repository and the chosen transport mode. This then
controls the transport flow of a dedicated component version.

Additionally it should also handle the preserving of existing artifacts for overwrite transports.

There are several new tests now for those scenarios.
 
Fixes #513

**Release Notes**:

```feature user

```

### DIFF
--- a/pkg/contexts/ocm/compdesc/componentdescriptor.go
+++ b/pkg/contexts/ocm/compdesc/componentdescriptor.go
@@ -379,6 +379,11 @@ func GenericAccessSpec(un *runtime.UnstructuredTypedObject) AccessSpec {
 	}
 }
 
+// AccessProvider provides access to an access specification of elements.
+type AccessProvider interface {
+	GetAccess() AccessSpec
+}
+
 // Sources describes a set of source specifications.
 type Sources []Source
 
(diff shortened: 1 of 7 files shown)


morri-son (approved):
@mandelsoft, approved. Please check if the next integration test runs successfully through

PR #625, opened 2024-01-09
Make ocm flake.nix ready

vasu1124:
## Description

This is a flake enablement for the repository of ocm. nix enabled environments can easily import ocm with
```
nix run github:open-component-model/ocm
nix run github:open-component-model/ocm#demo
nix run github:open-component-model/ocm#helminstaller
...
```
nix can hermetically build the package automatically on various architectures. 
With the flakes includes the the ocm repo, we can contribute our project into the nixos.org package library.

- [x] 🍕 Feature

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 
   inputs = {
     # NixPkgs (nixos-23.11)
-    nixpkgs.url = "nixpkgs/nixos-23.11"; #"github:NixOS/nixpkgs/nixos-23.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
   };
 
   outputs = { self, nixpkgs, ... }:
@@ -64,15 +64,17 @@ SPDX-License-Identifier: Apache-2.0
             nativeBuildInputs = [ pkgs.installShellFiles ];
 
             postInstall = ''
-              installShellCompletion --cmd ${pname} --zsh  <($out/bin/${pname} completion zsh)
-              installShellCompletion --cmd ${pname} --bash <($out/bin/${pname} completion bash)
-              installShellCompletion --cmd ${pname} --fish <($out/bin/${pname} completion fish)
+              installShellCompletion --cmd ${pname} \
+                  --zsh  <($out/bin/${pname} completion zsh) \
+                  --bash <($out/bin/${pname} completion bash) \
+                  --fish <($out/bin/${pname} completion fish)
             '';
 
             meta = with lib; {
               description = "Open Component Model (OCM) is an open standard to describe software bills of delivery (SBOD)";
               longDescription = ''
-                The Open Component Model (OCM) is an open standard to describe software bills of delivery (SBOD). OCM is a technology-agnostic and machine-readable format focused on the software artifacts that must be delivered for software products.
+                OCM is a technology-agnostic and machine-readable format focused on the software artifacts that must be delivered for software products.
+                The specification is also used to express metadata needed for security, compliance, and certification purpose.
               '';
               homepage = "https://ocm.software";
               license = licenses.asl20;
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -0,0 +1,125 @@
...
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+{
+  description = "Nix flake for ocm";
+
+  inputs = {
+    # NixPkgs (nixos-23.11)
+    nixpkgs.url = "nixpkgs/nixos-23.11"; #"github:NixOS/nixpkgs/nixos-23.11";
```
SuperSandro2000, line 11:
```suggestion
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
```
you want to use github:nixos/.... and not rely on the nix registry as that might point to something else which would break the lock file for those users

flake.nix
```diff
@@ -0,0 +1,125 @@
...
+        in
+        {
+          ${pname} = pkgs.buildGo121Module rec {
+            inherit pname self;
+            version = lib.fileContents ./VERSION;
+            gitCommit = if (self ? rev) then self.rev else self.dirtyRev;
+            state = if (self ? rev) then "clean" else "dirty";
+
+            # This vendorHash represents a dervative of all go.mod dependancies and needs to be adjusted with every change
+            vendorHash = "sha256-CA3p9QNHo7mHCoXkuOojFAJen3TdieSsoQVivxPk3yw=";
```
SuperSandro2000, line 43: This is a bummer but doing vendoring with k8s is also not a great option.

If people would have configured the sandbox to relaxed, we could fetch that but that is also not great.

flake.nix
```diff
@@ -0,0 +1,125 @@
...
+              "cmds/demoplugin"
+              "cmds/ecrplugin"
+            ];
+
+            nativeBuildInputs = [ pkgs.installShellFiles ];
+
+            postInstall = ''
+              installShellCompletion --cmd ${pname} --zsh  <($out/bin/${pname} completion zsh)
+              installShellCompletion --cmd ${pname} --bash <($out/bin/${pname} completion bash)
+              installShellCompletion --cmd ${pname} --fish <($out/bin/${pname} completion fish)
```
SuperSandro2000, line 69:
```suggestion
              installShellCompletion --cmd ${pname} \
                --zsh  <($out/bin/${pname} completion zsh) \
                --bash <($out/bin/${pname} completion bash) \
                --fish <($out/bin/${pname} completion fish)
```

flake.nix
```diff
@@ -0,0 +1,125 @@
...
+            '';
+
+            meta = with lib; {
+              description = "Open Component Model (OCM) is an open standard to describe software bills of delivery (SBOD)";
+              longDescription = ''
+                The Open Component Model (OCM) is an open standard to describe software bills of delivery (SBOD). OCM is a technology-agnostic and machine-readable format focused on the software artifacts that must be delivered for software products.
```
SuperSandro2000, line 75:
```suggestion
                OCM is a technology-agnostic and machine-readable format focused on the software artifacts that must be delivered for software products.
```
Since this metadata is mostly for https://search.nixos.org/, where description and longDescription is shown next to each other, the text will be kinda duplicated.

flake.nix
```diff
@@ -0,0 +1,125 @@
...
+        let 
+          pkgs = nixpkgsFor.${system};
+        in
+        {
+          default = pkgs.mkShell {
+            buildInputs = with pkgs; [ 
+              go_1_21 
+              gopls 
+              gotools 
+              go-tools
```
SuperSandro2000, line 95:
```suggestion
              gotools # goimports
              go-tools # staticcheck
```
I would recommend to document what contains what 😅

flake.nix
vasu1124: done

flake.nix
vasu1124: done

flake.nix
vasu1124: done

flake.nix
vasu1124: didn't you mention that we could use this nix-update for that? let's take a look at that later ...

mandelsoft (approved):

PR #684, opened 2024-03-06
MIT LICENSE for npm publish gist

hilmarf:
From cloverstd <cloverstd@gmail.com>
Fr 16.02.2024 17:10

You can use the code of this gist under the MIT LICENSE.

fixes: view-source:https://api.reuse.software/sbom/github.com/open-component-model/ocm.spdx


morri-son (approved):
@hilmarf if the REUSE config does not work with lowering the scope through additional Files: config, then we may add the headers in those files explicitly and use the FilesExcluded instead (https://github.com/open-component-model/.github/blob/main/CONTRIBUTING.md#licensing-and-copyright-information-on-file-level)

PR #715, opened 2024-04-02
remove docs that are available in the spec

morri-son:
## Description

remove parts from the "docs" folder that have been moved to the ocm-spec


hilmarf:
@morri-son - would be good to have a link to the docs, where they are located now

hilmarf (approved):

morri-son:
Docs can be found in the com-spec now: 

https://github.com/open-component-model/ocm-spec/blob/main/README.md

PR #734, opened 2024-04-20
Fix ocm issue 179,  block in config yields invalid yaml

dee0sap:
## Description

Fixes https://github.com/open-component-model/ocm-project/issues/179

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Related Issue # 179
- Closes # (179)
- Fixes # (179)
> Remove if not applicable

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [x] 🙅 no documentation needed

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


Skarlso:
@dee0sap All you have to do is indent the value that you wish to substitute correctly.

We marshal values before doing anything this PR now fixes that in the sense that we don't marshal primitive values. That marshalling broke the syntax of the value that we are trying to insert. So we leave values alone. But we do marshal complex types like struct, pointer, maps, etc.

pkg/utils/subst/subst.go
```diff
@@ -36,110 +42,156 @@ func ParseFile(file string, fss ...vfs.FileSystem) (SubstitutionTarget, error) {
 	return s, nil
 }
 
+var yqLibLogInit sync.Once
+
 func Parse(data []byte) (SubstitutionTarget, error) {
+	yqLibLogInit.Do(func() {
+		var lvl glog.Level
+		switch ocmlog.Context().GetDefaultLevel() {
```
Skarlso, line 50: I'm not sure this is needed honestly. If you mock with the library you won't change the default logging level on the package side. You would change it for the context logger that your are using throughout your calls. For which this would have no effect at all.

I would just disable the logging completely for yq.

Alternatively, modify the parser to take an OCM context or a logger context that could be passed in and use that. Something like, WithLogger, or something. I doubt too many things use this as is, so it could be a fairly okay breaking change.

pkg/utils/subst/subst.go
```diff
@@ -36,110 +42,156 @@ func ParseFile(file string, fss ...vfs.FileSystem) (SubstitutionTarget, error) {
...
-	// mixed json/yaml cannot be parsed, modified and marshalled again, correctly
-	// so try to come with pure yaml or pure json.
 
-	fi.content, err = parser.ParseBytes(data, 0)
-	if err != nil {
-		return nil, errors.Wrapf(err, "invalid YAML")
-	}
+	fi.content.SetDocument(0)
+	fi.content.SetFilename("substitution-target")
+	fi.content.SetFileIndex(0)
```
Skarlso, line 95: Why not simply use `NewAllAtOnceEvaluator` and evulate the nodes instead of doing SetDocument and the likes? I think we can completely remove the need for using / passing around the AST. We can just take in something and actually produce the output as a `[]byte`.

pkg/utils/subst/subst.go
```diff
@@ -36,110 +42,156 @@ func ParseFile(file string, fss ...vfs.FileSystem) (SubstitutionTarget, error) {
...
 	}
+
+	if bval, err := mrshl(value); err != nil {
+		return err
+	} else {
+		return f.SubstituteByData(path, bval)
+	}
+}
+
+func (f *fileinfo) substituteByValue(path string, value *yqlib.CandidateNode) error {
```
Skarlso, line 176: Is there a reason for using all these substitution parsings instead of a `NewAllAtOnceEvaluator` that produces a `[]byte` result given a printer?

pkg/utils/subst/subst_test.go
```diff
@@ -5,9 +5,6 @@
 package subst
 
 import (
-	"fmt"
```
Skarlso, line 8: Also note that the tests are using SubstituteByData and SubstituteByValue interchangingly and sometimes they do the same thing and sometimes they don't. Which is super confusing. We should provide a clear way where you can use a struct or not or provide a single function that marshalls a struct if it detects that the given object is not a string or a byte slice.

pkg/utils/subst/subst.go
dee0sap: I definitely don't want to disable the logging completely.   While it was chatty prior to this change it also turned out useful a couple of times.    Saved me from having to step through everything to understand what went wrong.

It does make sense to pass in the Context, however I really think that should be a separate PR.

Since the mapping would be needed once Context is passed in, I think my opinion is that 
- this mapping logic remain as is for now
- a ticket be created for changing the interface to take a Context 
- in the above ticket, mention that the mapping should be changed to map from the passed in Context when it is available

pkg/utils/subst/subst.go
dee0sap: I wouldn't want to remove the caching of the AST in this PR for fear of having a negative impact on performance.  I say this because 
- I see ocm-controller caching SubstitutionTargets during mutation ( so someone has some perf concerns already )
- Similarly SubstitutionTargets were caching the AST 
- I know that in CSI we will be making several updates to the same SubstitutionTarget, so if parsing and reparsing is slow it will impact us
- We're already doing 1 thing that could have a performance impact, changing the lib, and if we make another change, taking out the cache, and someone complains we won't know which change was the problem
- I haven't seen any performance tests that would tell us that removing the AST caching is safe

So again, in my opinion if taking out the caching is desirable for some reason I think that should be a separate PR and that PR should also add some performance tests.

Oh, and a non-performance concern.... not having the cached object makes debugging more difficult if you are having multiple updates happening to the same substitution target because in `Substitute(subs Substitutions, fs vfs.FileSystem) ` your won't be able to 
- You won't be able to add a watch that lets you quickly see what each of the rules are doing to the SubstitutionTarget
- You won't be able to add a conditional breakpoint that lets you catch which rule is making a particular change to a SubstitutionTarget

pkg/utils/subst/subst.go
dee0sap: - As mentioned above, I wouldn't want to switch the code to hang on to a byte array instead of the AST
- The combo of NewAllAtOnceEvaluator the returned Evaluator interface don't let me set the Context and that is how we pass the substitution value to yqlib

pkg/utils/subst/subst_test.go
dee0sap: Not sure exactly what you're saying here.    I see that the SubstitutionTarget interface provides these methods
- SubstituteByData(path string, value []byte) error
- SubstituteByValue(path string, value interface{}) error
and I see there are some tests that test one and some tests that test the other.   So long as the interface is providing those two methods it makes sense to me that some tests are testing the one method while some tests are testing the other.

If you are suggesting that the SubstitutionTarget only have 1 method, then I really don't that change should be part of this PR.

pkg/utils/subst/subst.go
Skarlso: Fair point 👍

pkg/utils/subst/subst.go
Skarlso: 👍

pkg/utils/subst/subst_test.go
Skarlso: True. It's not part of this pr.

pkg/utils/subst/subst.go
Skarlso: Right, but those are for subs rules. That would happen once with the other parser as well. 🤔

Anyways, I played around with it a bit and actually don't really like it, because it fucks up the multiline formatting. It adds that pesky little `|-` to everything. :D And it's so annoying. So I don't care. :D Leave it be like that. :)

pkg/utils/subst/subst.go
Skarlso: 👍

pkg/utils/subst/subst.go
Skarlso: Also, the JSON part is completely broken somehow and it further ignores a couple of proper indentations. It's really weird, that it works nicely, once you get into AST level. :D So I think what you are doing is better.

Skarlso (approved):
LGTM

Skarlso:
If anything breaks because of this, we'll iterate on a solution. ;)

PR #754, opened 2024-04-29
it's not that hard to do proper logging... is it?

hilmarf:
@Skarlso - wouldn't that be an easy solution to simple log, where it's not possible to return errors?


Skarlso:
>it's not that hard to do proper logging... is it?

The logger comes from OCM context. This is an uninitialized logger that logs to god knows what. You might as well use fmt.Println at that point.

pkg/logging/logError.go
```diff
@@ -0,0 +1,7 @@
+package logging
+
+func Error(err error, keypairs ...interface{}) {
+	if err != nil {
+		Context().Logger(DefineSubRealm("error")).Error(err.Error(), keypairs...)
```
Skarlso, line 5: The context comes from the library / CLI. The problem with doing this is that it completely ignores what the context setup is. It could be the CLI which logs to stdout or error or the Library where the logger will log to a configured medium that is not represented here.

Either way, this will not work as the context is not passed into these situations so it will log to an arbitrary location. Worst case, it displays a warning saying something like this

```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.

```

hilmarf: argh... and a dump and simple: `println()` ?

Skarlso: That still could produce an unexpected output to the wrong logging medium. :(

PR #757, opened 2024-05-02
Npm/auth access

### LINKED ISSUE #753: Credentials for npm repositories have no effect
**What happened**:
Credentials for npm are configured as documented in .ocmconfig or command line:
```
type: generic.config.ocm.software/v1
configurations:
  - type: credentials.config.ocm.software
    consumers:
      - identity:
          type: Registry.npmjs.com
          hostname: common.repositories.cloud.sap
        credentials:
          - type: Credentials
            properties:
              username: gardener-lm-ocm
              password: ...
```

Access to private npm registries is possible. But results in 401 http error

any resource with type: npmPackage and located in any private npm registry

The type ` type: Registry.npmjs.com`differs in syntax from all other types (OCIRegisrty, HelmChartRepository) and why a capital R?

**Environment**:

hilmarf:
## Description

When credentials are configured, let's also use them for the read access.

- [x] 🍕 Feature
- [ ] 🎇 Restructuring
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

- Closes #753

## Added tests?

- [ ] 👍 yes

## Added to documentation?

- [ ] 📜 README.md
- [ ] 🙅 no documentation needed

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules


pkg/contexts/credentials/builtin/npm/identity/identity.go
```diff
@@ -65,3 +72,86 @@ func GetCredentials(ctx cpi.ContextProvider, repoUrl string, pkgName string) com
 	}
 	return credentials.Properties()
 }
+
+// BearerToken retrieves the bearer token for the given repository URL and package name.
+// Either it's setup in the credentials or it will login to the registry and retrieve it.
+func BearerToken(ctx cpi.ContextProvider, repoUrl string, pkgName string) (string, error) {
```
mandelsoft, line 78: Those functions are typically in a technology support package (like pkg/helm). The identity packages are typically just used to describe the technology specific identity mapping for the uniform credential handling mechanism, so they deal with the identity and credential attributes and identity matchers used for the credential mechanism to find and describe credential sets. If you want to put technology specific access code into this identity package you should at least use different file names, not the standard names used for the mapping in the other packages to keep some kind of consistency. Better would be to put technology specific access code into a separate package to be used by access methods, blob access and uploader packages.

Basically all identity packages have some GetCredentials function taking the used identity parameters and providing found credential properties in a standard way. This method is then used in the technology support package to do technology specific things like you did in this method.

pkg/contexts/ocm/accessmethods/npm/method.go
```diff
@@ -96,11 +94,11 @@ func (a *AccessSpec) GetInexpensiveContentVersionIdentity(access accspeccpi.Comp
 }
 
 func (a *AccessSpec) getPackageMeta(ctx accspeccpi.Context) (*meta, error) {
-	url := a.Registry + path.Join("/", a.Package, a.Version)
```
mandelsoft, line 99: you moved this code in this method to change the reader method, but copied it into the reader method, also.

I think, this is, because you need parts of the attributes incorporated into the url as dedicated fields.
May be you should then add a method on the AccessSpec to calculate the URL to avoid the code duplication.

pkg/contexts/credentials/builtin/npm/identity/identity.go
```diff
@@ -65,3 +68,47 @@ func GetCredentials(ctx cpi.ContextProvider, repoUrl string, pkgName string) com
...
```
fabianburth, line 74: I think to stay uniform with the existing packages, we should also move `BearerToken()` and `Authorize()` to the npm login file.

fabianburth (changes requested):

pkg/contexts/credentials/builtin/npm/identity/identity.go
hilmarf: but that would lead to cyclic dependencies between the `npm.identity` and the generic npm package

hilmarf: done

pkg/contexts/ocm/accessmethods/npm/method.go
hilmarf: done

pkg/contexts/credentials/builtin/npm/identity/identity.go
hilmarf: done

fabianburth (approved):

PR #791, opened 2024-06-03
fix update-vendor-hash

hilmarf:
The current GHA fails, because we require go v1.22.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,7 @@
 
   inputs = {
     # NixPkgs (nixos-23.11)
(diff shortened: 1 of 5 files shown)


flake.nix
vasu1124, line 1: /lgtm

vasu1124 (approved):

Skarlso (dismissed):

Skarlso (dismissed):

Skarlso:
I knew this would happen. :P

Skarlso:
Hmmm.. instead of reverting, we should look into fixing those, don't we?

Skarlso (approved):

PR #810, opened 2024-06-10
Fix make cmds

fabianburth:
This PR adds `gci` to `make install-requirements`. This fixes a bug in the `gci write` commands which lead to the `slices` import being wrongly sorted which made `make format` and also `make prepare` essentially unusable. 

During the adjustment, I also changed `make test-all` to `make test` and the previous `make test` to `make unit-test`. I think the default should be to run all tests (currently, they dont take too long anyway).


hack/Makefile
```diff
@@ -33,7 +33,7 @@ ifeq ("v$(JQ)","v0.0")
 	deps += $(detected_OS)_jq
 endif
 
-GOLANGCILINT_VERSION := "v1.54.2"
+GOLANGCILINT_VERSION := "v1.59.1"
```
Skarlso, line 36: This will result in a f*ckton of errors. :D Just FYI :D

Makefile
```diff
@@ -54,6 +54,10 @@ format:
 check:
 	@$(REPO_ROOT)/hack/check.sh --golangci-lint-config=./.golangci.yaml $(REPO_ROOT)/cmds/ocm $(REPO_ROOT)/cmds/helminstaller/... $(REPO_ROOT)/cmds/ecrplugin/... $(REPO_ROOT)/cmds/demoplugin/... $(REPO_ROOT)/pkg/...
 
+.PHONY: check-and-fix
+check-with-fix:
+	@$(REPO_ROOT)/hack/check.sh --fix --golangci-lint-config=./.golangci.yaml $(REPO_ROOT)/cmds/ocm $(REPO_ROOT)/cmds/helminstaller/... $(REPO_ROOT)/cmds/ecrplugin/... $(REPO_ROOT)/cmds/demoplugin/... $(REPO_ROOT)/pkg/...
```
hilmarf, line 59: why aren't the three dots (`/...`) at ocm command required?

```suggestion
	@$(REPO_ROOT)/hack/check.sh --fix --golangci-lint-config=./.golangci.yaml $(REPO_ROOT)/cmds/ocm/... $(REPO_ROOT)/cmds/helminstaller/... $(REPO_ROOT)/cmds/ecrplugin/... $(REPO_ROOT)/cmds/demoplugin/... $(REPO_ROOT)/pkg/...
```

fabianburth: Good catch! Those files have essentially never been linted. Had to fix quite some errors.

cmds/ocm/pkg/processing/logging.go
```diff
@@ -6,4 +6,4 @@ import (
 
 var REALM = ocmlog.DefineSubRealm("output processing chains", "processing")
 
-var logger = ocmlog.DynamicLogger()
+// var logger = ocmlog.DynamicLogger()
```
hilmarf, line 9: shouldn't it be removed completely?

fabianburth: Done

hilmarf (approved):

mandelsoft (dismissed):

mandelsoft (approved):

PR #826, opened 2024-06-24
enhance the auto update of the flake vendor hash

hilmarf:

- create the commit only if there is a difference
- create commit on the branch, which change go.mod/go.sum
- act on pull requests


morri-son (approved):
lgtm

PR #844, opened 2024-07-14
Rework adding elements to a component version

### LINKED ISSUE #842: Unexpectedly added `version` property to `extraIdentity`
**What happened**:
When executing `ocm sign componentversions` on a component descriptor which contains multiple resources which share the same name and have `extraIdentity` explicitly set to `{}`, all resources have the `version` property added to the `extraIdentity` field except the last resource of that name (which is reasonable since the last resource is unique without `extraIdentity` being set as soon as all other resources have their `version` property added to it).

Don't silently add properties to the `extraIdentity` field during `ocm sign` command (and certainly not inconsistently). Instead, rather fail verification.

Example component descriptor which has to be signed using `ocm sign componentversions` command:
```
component:
  componentReferences: []
  name: example.org/my-component
  provider: ACME Inc.
  repositoryContexts: []
  resources:
  - access:
      imageReference: hello-world
      type: ociArtifact
    extraIdentity: {} # must not be omitted to reproduce behaviour
    name: my-resource # must be part of the component multiple times
    relation: external
    type: ociImage
    version: 0.1.1 # must be different for the resources sharing the same name
  - access:
      imageReference: hello-world
      type: ociArtifact
    extraIdentity: {} # must not be omitted to reproduce behaviour
    name: my-resource # must be part of the component multiple times
    relation: external
    type: ociImage
    version: 0.1.2 # must be different for the resources sharing the same name
  sources: []
  version: 1.0.0
meta:
  schemaVersion: v2
signatures: []
```

**Environment**:

mandelsoft:

The previous mechanism for adding elements (resources, sources and references) to a component version
provided a hybrid mechanism, only:
- check if an element with the given identity already exists -> replace it
- if no entry is found -> append

This behaviour does not allow to add elements with the same base identity but a different version, if the version
is not explicitly added to the extra identity.
So, far this explicit handling of the version is not required, because it is implicitly done when evaluating the element list.

This PR add new options to the Set methods for the three element kinds on the component version access to specify the exact desired behaviour:
- Replace a dedicated index
- Replace a dedicated identity
- Update an existing entry
- Append a new entry

The default behaviour is unchanged. It is sufficient for the most common use cases, as long as no
duplicate entries with different versions should be added.

The CLI commands have been adjusted to use the *Append* mode by default. This is exactly what is desired if new  component versions are composed.
If an existing version should be modified, it is possible to use the option `-R` to switch to the default mode.

Fixes #842


cmds/ocm/commands/ocmcmds/common/addhdlrs/base.go
```diff
@@ -0,0 +1,39 @@
+package addhdlrs
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/open-component-model/ocm/cmds/ocm/pkg/options"
+	"github.com/open-component-model/ocm/pkg/contexts/ocm"
+)
+
+type ResourceSpecHandlerBase struct {
```
fabianburth, line 10: Shouldn't this thing be called `ElementSpecHandlerBase`?

fabianburth (changes requested):

fabianburth (approved):

frewilhelm (approved):

PR #850, opened 2024-07-22
Improve errors

fabianburth:
`ocm transfer cv`  command ignored the -t flag if the ctf already existed. This PR adds documentation to specify that this is intended behavior.

The ocm download command `ocm download cv` command did not return anything (not even an error) if the requested component version did not exist. This is fixed by this PR. 

Fixes https://github.com/open-component-model/ocm-project/issues/203

The nil pointer exception was already fixed with version v0.12.0. But as mentioned above, users got no feedback.


frewilhelm (approved):
lgtm

PR #855, opened 2024-07-27
Add AUR as installation option for OCM-CLI

ccwienk:
Make OCM more convenient to consume for users of Arch Linux.

Note: there are different ways (and tools) to consume packages from AUR (hence just a reference to documentation rather than a concrete shellscript-snippet). I think it is safe to assume Arch Linux users will be able to consume packages from AUR without such help ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Install the latest release from any of
 
 - [Homebrew](https://brew.sh)
 - [Nix](https://nixos.org)
+- [AUR](https://aur.archlinux.org/packages/ocm-cli)
 - [Docker](https://www.docker.com/)
 - [Podman](https://podman.io/)
 - [GitHub Releases](https://github.com/open-component-model/ocm/releases)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,6 +93,12 @@ nix profile list | grep ocm
 ocm --help
 ```
 
+### Install from AUR (Arch Linux User Repository)
+
+git-urL: https://aur.archlinux.org/ocm-cli.git
+
+[AUR Documentation](https://wiki.archlinux.org/title/Arch_User_Repository)
+
```
vasu1124, line 101: can you please document the rudimentary how to steps on installing the package? referencing to the AUR Documentation is in imho good, but not sufficient.
also, i found this link https://aur.archlinux.org/packages/ocm-cli that also includes a reference to which version is current.

ccwienk: @vasu1124 : done
as for the page-link: I added this earlier above (line 59). I can of course repeat it here 🤷🏻‍♂️ 
-> done

ccwienk: on another note: if someone would be interested to include updating into release-pipeline, I can offer to grant required privileges to pipeline-user. I already created a shellscript for doing the updateing, so integration should be quite easy

morri-son: For sure we should have the upload of the latest released package to all mentioned package managers in the Github action "Release". I opened https://github.com/open-component-model/ocm/issues/856 to track that.

morri-son (approved):

morri-son:
@ccwienk I saw that we also have the installation part mentioned in the "getting started" guide. Just referencing the Readme doc in the ocm repo would be the best idea to avoid double-maintenance. But since the site/doc structure on the website slightly differs and the page should not just contain a single link, I now maintained it there, too: https://github.com/open-component-model/ocm-website/pull/302. Not ideal, I know, but before we have a final concept where and how to create and store the content that should be presented on the website, this is the only way...

PR #865, opened 2024-08-06
Resource resolution on CD set

mandelsoft:

Introduction resolution of relative resource references of the level of component descriptors.

So far, this the resolution was only possible on the level of the complete component version access abstraction.
This PR adds such a functionality on the level of plain component descriptors.
Therefore, it provides a `ComponentVersionResolver` abstraction for component descriptors, with two
implementations: a set of component descriptors and a compound resolver similar to the one for component version accesses.

To align the element names in package `compdesc`, the `ComponentReference`ans been renamed to `Reference` (like `Resource` and `Source`). The old type has been deprecated.

This functionality is required for the OCM controllers working an deep resources for a component version.

To verify the digest of the retrieved content of a resource, a new function
`signing.VerifyResourceDigest(cv, index, content)`
is provided. it uses the digesting type described by the component version to recalculate the digest of the retrieved resource and verify it against the digest described by the component version.

### DIFF
--- a/api/ocm/compdesc/meta/v1/identity.go
+++ b/api/ocm/compdesc/meta/v1/identity.go
@@ -103,6 +103,22 @@ func (i Identity) Remove(name string) bool {
 	return false
 }
 
+// ExtraIdentity extracts the extra identity part of an identity.
+func (i Identity) ExtraIdentity() Identity {
+	if i == nil {
+		return nil
+	}
+	if _, ok := i[SystemIdentityName]; !ok {
+		return i
+	}
+	if len(i) == 1 {
+		return nil
+	}
+	r := i.Copy()
+	delete(r, SystemIdentityName)
+	return r
+}
+
 func (i Identity) String() string {
 	if i == nil {
 		return ""
--- a/api/ocm/compdesc/resolver.go
+++ b/api/ocm/compdesc/resolver.go
@@ -92,5 +92,8 @@ func (c *CompoundResolver) LookupComponentVersion(name string, version string) (
 func (c *CompoundResolver) AddResolver(r ComponentVersionResolver) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
-	c.resolvers = append(c.resolvers, r)
+
+	if r != nil {
+		c.resolvers = append(c.resolvers, r)
+	}
 }
--- a/api/ocm/compdesc/resolver.go
+++ b/api/ocm/compdesc/resolver.go
@@ -1,0 +1,96 @@
+package compdesc
+
+import (
+	"sync"
+
+	"github.com/mandelsoft/goutils/errors"
+
+	common "ocm.software/ocm/api/utils/misc"
+)
+
+type ComponentVersionResolver interface {
+	LookupComponentVersion(name string, version string) (*ComponentDescriptor, error)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+type ComponentVersionSet struct {
+	lock sync.RWMutex
+	cds  map[common.NameVersion]*ComponentDescriptor
+}
+
+var _ ComponentVersionResolver = (*ComponentVersionSet)(nil)
+
+func NewComponentVersionSet(cds ...*ComponentDescriptor) *ComponentVersionSet {
+	r := map[common.NameVersion]*ComponentDescriptor{}
+	for _, cd := range cds {
+		r[common.NewNameVersion(cd.Name, cd.Version)] = cd.Copy()
+	}
+	return &ComponentVersionSet{cds: r}
+}
+
+func (c *ComponentVersionSet) LookupComponentVersion(name string, version string) (*ComponentDescriptor, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	nv := common.NewNameVersion(name, version)
+
+	cd := c.cds[nv]
+	if cd == nil {
+		return nil, errors.ErrNotFound(KIND_COMPONENTVERSION, nv.String())
+	}
+	return cd, nil
+}
+
+func (c *ComponentVersionSet) AddVersion(cd *ComponentDescriptor) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.cds[common.NewNameVersion(cd.Name, cd.Version)] = cd.Copy()
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+type CompoundResolver struct {
+	lock      sync.RWMutex
+	resolvers []ComponentVersionResolver
+}
+
+var _ ComponentVersionResolver = (*CompoundResolver)(nil)
+
+func NewCompoundResolver(res ...ComponentVersionResolver) ComponentVersionResolver {
+	for i := 0; i < len(res); i++ {
+		if res[i] == nil {
+			res = append(res[:i], res[i+1:]...)
+			i--
+		}
+	}
+	if len(res) == 1 {
+		return res[0]
+	}
+	return &CompoundResolver{resolvers: res}
+}
+
+func (c *CompoundResolver) LookupComponentVersion(name string, version string) (*ComponentDescriptor, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	for _, r := range c.resolvers {
+		if r == nil {
+			continue
+		}
+		cv, err := r.LookupComponentVersion(name, version)
+		if err == nil && cv != nil {
+			return cv, nil
+		}
+		if !errors.IsErrNotFoundKind(err, KIND_COMPONENTVERSION) && !errors.IsErrNotFoundKind(err, KIND_COMPONENT) {
+			return nil, err
+		}
+	}
+	return nil, errors.ErrNotFound(KIND_REFERENCE, common.NewNameVersion(name, version).String())
+}
+
+func (c *CompoundResolver) AddResolver(r ComponentVersionResolver) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.resolvers = append(c.resolvers, r)
+}
--- a/api/ocm/tools/signing/store.go
+++ b/api/ocm/tools/signing/store.go
@@ -81,7 +81,7 @@ func (v *verifiedStore) Load() error {
 	var storage StorageDescriptor
 	f, err := v.fs.Open(v.file)
 	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
+		if !errors.Is(err, os.ErrNotExist) {
 			return err
 		}
 	} else {
--- a/api/ocm/tools/signing/store.go
+++ b/api/ocm/tools/signing/store.go
@@ -1,0 +1,192 @@
+package signing
+
+import (
+	"io"
+	"os"
+	"sync"
+
+	"github.com/mandelsoft/filepath/pkg/filepath"
+	"github.com/mandelsoft/goutils/errors"
+	"github.com/mandelsoft/goutils/sliceutils"
+	"github.com/mandelsoft/vfs/pkg/vfs"
+
+	"ocm.software/ocm/api/ocm/compdesc"
+	metav1 "ocm.software/ocm/api/ocm/compdesc/meta/v1"
+	"ocm.software/ocm/api/utils"
+	common "ocm.software/ocm/api/utils/misc"
+	"ocm.software/ocm/api/utils/runtime"
+)
+
+type VerifiedStore interface {
+	Add(cd *compdesc.ComponentDescriptor, signatures ...string)
+	Remove(n common.VersionedElement)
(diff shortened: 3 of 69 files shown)


Skarlso (approved):

Skarlso:
So the current method of recursive lookup for a Resource is simply doing this:

```go
	identities = append(identities, resource.ReferencePath...)

	res, _, err := utils.ResolveResourceReference(
		cva,
		ocmmetav1.NewNestedResourceRef(ocmmetav1.NewIdentity(resource.Name), identities),
		cva.Repository(),
	)
	if err != nil {
		return nil, "", -1, fmt.Errorf(
			"failed to resolve reference path to resource: %s %w",
			resource.Name,
			err,
		)
	}
```

Given an identity with the referencePath and the root component that function just simply figured out where the resource is given the path.

With this approach, and please correct me if I'm wrong, we have to construct a Set ( which needs all of the component descriptors from all the nested references), then a resolver for that set then call ResolveResourcePath and get our resource. Is that correct?

mandelsoft:
Yes, there is basically just a second resolution environment for component descriptors similar to the already existing one working on `ocm.ComponentVersionAccess` objects and ocm.ComponentVersionResolver`s.

This environment is in package `ocm.compdesc`:
- A special descriptor resolver interface `ComponentVersionResolver`
- A `ComponentVersionSet`  as standard implementation for this interface, hosting just a set of `ComponentDescriptor`s.
- A `CompoundResolver` to bundle multiple resolvers
- and the approriate functions to resolve references: `ResolveReferencePath` and `ResolveResourceReference`

These elements are 1:1 mappings of the current ones (in `ocmutils`)  working on `ComponentVersionAccess`es.

Both environments work on the standard `etav1.ResourceReference`. But the compdesc version provide the `Resource` entry and the `ComponentDescriptor` of the requested resource. To access the resource content. The om library can be used directly of the component version of the found component descriptor.

Example:
```
  set := compdesc.NewComponentVersionSet()
  set.AddVersion(cd1)
  set.AddVersion(cd2)
  ...

 refpath := []metav1.Identity{...}
 rscid := metav1.NewIdentity("myresource")

  ref := metav1.NewNestedResourceRef(rscid, refpath)

or

  ref := metav1.NewResourceRef(rscid, comprefid1, ....)

 rsc, effcd, err := compdesc.ResolveResourceReference(rootcd, ref, set)
  ...
  
.
  var ocmrepo Repository = ...

  cv, err := ocmrepo.LookupComponentVersion(effcd.GetName(). effcd.getVersion())

  reader, err := ocmutils.GetResourceReaderForPath(cv, ref.Resource)

or
  r, err := cv.GetResource(iref.Resource)
  reader, err :=  GetResourceReader(r)
		
```

mandelsoft:
I added a new function`signing.VerifyResourceDigest(cv, index, content)` to do this job,

Skarlso:
Okay, sure, however, my question was, that we already have a utils method that resolves and finds the right resource given an extra identity component path and the root component without the need for the resolvers and the set and the extra implementation of finding all of the component versions and putting them into the set.

So I was wondering what is the benefit of this implementation over the simple `utils.ResolveResourceReference`?

Is it that if we parse all of the `ComponentDescriptors` into k8s objects, we can provide those k8s objects to this set / resolver combination and it doesn't matter in what order we'll do that this thing will just figure it out which Component Version the searched Resource resides in?

mandelsoft:
@Skarlso , exactly, if the k8s resource `Component` verifies the component version it should expose the transitive closure of component descriptors (which are covered by the verification). A k8s `Resource` resource is then able to use this set to resolve the resource reference to figure out the effective component version. If the digest found for this version matches the digest provided by the `Component` resource it can verify the downloaded artifact content.

Skarlso:
Gotcha, that makes sense. Thanks for the clarification.

api/ocm/compdesc/meta/v1/identity.go
```diff
@@ -103,6 +103,22 @@ func (i Identity) Remove(name string) bool {
...
+func (i Identity) ExtraIdentity() Identity {
+	if i == nil {
+		return nil
+	}
+	if _, ok := i[SystemIdentityName]; !ok {
+		return i
+	}
+	if len(i) == 1 {
+		return nil
+	}
```
Skarlso, line 116: Can you explain the logic here please?

If there is no system identity, then just returning the identity is a false positive, because that's not the ExtraIdentity that's just the identity, isn't it? Does this mean that if there is no system identity, the rest is just extra identity?

So anything else that isn't the system identity, is the extra identity, but why is it a problem if it's just one thing?

api/ocm/compdesc/meta/v1/identity.go
```diff
@@ -103,6 +103,22 @@ func (i Identity) Remove(name string) bool {
...
+	r := i.Copy()
+	delete(r, SystemIdentityName)
```
Skarlso, line 118: These two operations together cost more than just simply looping through the map and creating a new one but skipping the system identity key.

api/ocm/compdesc/resolver.go
```diff
@@ -0,0 +1,96 @@
...
+var _ ComponentVersionResolver = (*CompoundResolver)(nil)
+
+func NewCompoundResolver(res ...ComponentVersionResolver) ComponentVersionResolver {
+	for i := 0; i < len(res); i++ {
+		if res[i] == nil {
+			res = append(res[:i], res[i+1:]...)
+			i--
+		}
+	}
+	if len(res) == 1 {
```
Skarlso, line 68: Are you anticipating that someone might initialize this resolver with `(res, res, nil, nil, res, res)`? How would that be possible, or why would that be possible?

api/ocm/tools/signing/store.go
```diff
@@ -0,0 +1,211 @@
...
+	"golang.org/x/exp/maps"
+
+	"ocm.software/ocm/api/ocm/compdesc"
+	metav1 "ocm.software/ocm/api/ocm/compdesc/meta/v1"
+	"ocm.software/ocm/api/utils"
+	common "ocm.software/ocm/api/utils/misc"
+	"ocm.software/ocm/api/utils/runtime"
+)
+
+type VerifiedStore interface {
```
Skarlso, line 21: Please add some comments to these and explain what these new store things are.

api/ocm/tools/signing/store.go
```diff
@@ -0,0 +1,211 @@
...
+		return errors.ErrNotFound("directory", dir)
+	}
+
+	var storage StorageDescriptor
+	f, err := v.fs.Open(v.file)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+	} else {
```
Skarlso, line 91: Huh, why is it okay if it does not exist?

api/ocm/tools/signing/store.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil
+	}
+	r, err := cd.GetResourceByIdentity(id)
+	if err != nil {
+		return nil
+	}
+	return r.Digest
+}
+
+func (v *verifiedStore) GetResourceDigestByIndex(n common.VersionedElement, idx int) *metav1.DigestSpec {
```
Skarlso, line 193: Shouldn't this return an error when things aren't found or the index is out of range?

docs/reference/ocm_download_cli.md
```diff
@@ -91,6 +94,15 @@ This is handled by download handler. Their usage can be enabled with the <code>-
 option. Otherwise the resource as returned by the access method is stored.
 
 
+If the verification store is enabled, resources downloaded from
```
Skarlso, line 97: How is this functionality related in any way to the word "store"? I would assume something like `--verify` or something would be more akin to this functionality. Store, means storing something somewhere. This implies that it's just verification of some sort. And if it IS stored, then it should be explained where, for how long, why, and it what format and in under what circumstances the store is invalidated.

api/ocm/compdesc/meta/v1/identity.go
mandelsoft: The identity type may represent any set of identity attributes, not necessarily a complete identity.
The method ExtraIdentity just filters the attributes, which would be part of the extra identity of a complete element identity. Meaning: remove the SystemIdentityName attribute if set. If only this attribute is set, it would have no extra identity.

api/ocm/compdesc/meta/v1/identity.go
mandelsoft: yes, may be, but you have to implement it completely on your side, by copying the copy code, to be performant.

api/ocm/compdesc/meta/v1/identity.go
Skarlso: Okay, but if the systemIdentity is a part of the extra identity, why is it filtered out? And if it IS filtered out, why is it part of the extra identity? 🤔

api/ocm/compdesc/resolver.go
mandelsoft: the resolver list is typically variadic. On other places you may pass a resolver, which could be nil.
Therefore it seems to be useful to filter the nil at the final place , nit at all places where it is called to avoid always code
like 
```
   if resolver == nil {
       call(...)
   }  else {
       call(..., resolver)
   }
```

And generally this is some kind of defensive programming.

api/ocm/tools/signing/store.go
mandelsoft: The store must does not exist, when it is used for the first time. In this case only the folder must exists.
If it is nmodified later on, it saved on-the-fly, so that the next time it is used the correct content is available.

Otherwise the user always would have to create a file with the empty json/yaml representation. (potentially an empty file), before he could use the store option.

api/ocm/compdesc/resolver.go
Skarlso: But if you don't pass in anything the slice is just plain empty so nothing would happen. You would have to explicitly pass in a `nil` for that to be there something or something kind of list where there are nils in-between because the resolver was somehow not initialized.

And yes, I agree that you need to do the defence, but you are actually already doing that further down when you are looping through these resolvers here:

```go
	for _, r := range c.resolvers {
		if r == nil {
			continue
		}
```

So I think this looping here isn't really necessary.

docs/reference/ocm_download_cli.md
mandelsoft: Indeed, the store stores the verified component versions together with the verified digests, therefore store seems to be applicable.  If a CV is verified and not changed then it is verified forever. To invalidate your current verifications you just need to remove the store file or just use another one.

api/ocm/tools/signing/store.go
mandelsoft: yes I'll add this tomorrow.

docs/reference/ocm_download_cli.md
Skarlso: Ah, I see. I might have misunderstood the functionality of the store then. Thanks.

api/ocm/tools/signing/store.go
Skarlso: Gotcha, thanks for the explanation.

api/ocm/tools/signing/store.go
Skarlso: Gotcha, thanks!

api/ocm/compdesc/meta/v1/identity.go
mandelsoft: The extra identity is the content of the field `extraIdentity` plus optionally the `version``, but it never includes the `name` attribute. In the future we will require the `version`to be explicitly added to the `extraIdentity`field if it should be used, instead of implicitly adding if required to be unique.

So, formally, the identity of an element is the `name`of the element plut the attributed described by the `extraIdentity` field.  And vice versa, the extra identity of an element is the identity without the element name.

api/ocm/compdesc/resolver.go
mandelsoft: yes, sure, but when creating a compound resolver with one effective sub resolver the useless compound feature can be omitted, instead, the sole resolver is returned.
This requires to eliminate the nil resolvers first. What could be done is to omit the `continue` case in the Lookup method, if really all constructors assure that there is no nil entry. So far, this is the case, but possible constructors can be added later without knowing this special constraint, when just keep this check in the Lookup implementation.

mandelsoft: By the way, there is already such a case in the actual coding.
The AddResolver does not check for nil.
This should be fixed.

mandelsoft: I've added the check

Skarlso: Nice.

Skarlso (approved):

fabianburth (approved):

fabianburth (approved):

PR #866, opened 2024-08-06
Validate method for oci/ocm repository specs

mandelsoft:

This PR introduces a `Validate` method for OCI and OCM repository specifications.

Its task is to validate the spec as well as the connectivity (and potentially the credentials), if possible.
Checking credentials makes only sense, if there is no possibility to use different credentials for different shards
of the repository (for example organizations in the GitHub OCI registry).


api/oci/extensions/repositories/ocireg/type.go
```diff
@@ -128,6 +130,18 @@ func (a *RepositorySpec) Repository(ctx cpi.Context, creds credentials.Credentia
...
 
+func (a *RepositorySpec) Validate(ctx cpi.Context, creds credentials.Credentials, context ...credentials.UsageContext) error {
+	info, err := a.getInfo(creds)
+	if err != nil {
+		return err
+	}
+	h, p, _ := info.HostInfo()
+	if p == "" {
+		p = "443"
+	}
```
hilmarf, line 141: should we log here something like:
'No port provided falling back to default port 443'

api/oci/extensions/repositories/docker/docker_test.go
```diff
@@ -0,0 +1,20 @@
+//go:build docker_test
```
hilmarf, line 1: when and where is this test executed?

hilmarf (changes requested):
can we please have at least one place where the new 'Validate()' is go-documented?

api/oci/extensions/repositories/ocireg/type.go
mandelsoft: this is the default port used for OCI registry REST access. I think a log is not required, here

api/oci/extensions/repositories/docker/docker_test.go
mandelsoft: you can execute it locally. Docker is not necessarily available when executing the standard tests.
The same has been done for the redis pubsub extension.

hilmarf: okay, do we have an overview which additional integration tests are available?
are those already executed in https://github.com/open-component-model/ocm-integrationtest/ ?

hilmarf (approved):

hilmarf (approved):

Issue #864, opened 2024-08-05
unhandled nil-pointer-dereference (ocm list componentversions)


ccwienk:
**What happened**
```
$ ocm list componentversions example.com/my-component:1.0.0 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x2850cf2]

goroutine 1 [running]:
github.com/open-component-model/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr.(*TypeHandler).all(0xc00150a6c0, {0x0, 0x0})
	/home/runner/work/ocm/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr/typehandler.go:88 +0x32
github.com/open-component-model/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr.(*TypeHandler).get(0xc00150a6c0, {0x0, 0x0}, {0x3794780?, 0xc000417390?})
	/home/runner/work/ocm/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr/typehandler.go:213 +0x885
github.com/open-component-model/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr.(*TypeHandler).Get(0x37958c0?, {0x3794780?, 0xc000417390?})
	/home/runner/work/ocm/ocm/cmds/ocm/commands/ocmcmds/common/handlers/vershdlr/typehandler.go:109 +0x31
github.com/open-component-model/ocm/cmds/ocm/pkg/utils.HandleOutput({0x37bcba0, 0xc00002c5b0}, {0x37bdfe0, 0xc00150a6c0}, {0xc000417380?, 0x3?, 0xc00150a6c0?})
	/home/runner/work/ocm/ocm/cmds/ocm/pkg/utils/handling.go:84 +0x1c4
github.com/open-component-model/ocm/cmds/ocm/pkg/utils.HandleOutputs(...)
	/home/runner/work/ocm/ocm/cmds/ocm/pkg/utils/handling.go:64
github.com/open-component-model/ocm/cmds/ocm/pkg/utils.HandleArgs(0xc00008fea0, {0x37bdfe0, 0xc00150a6c0}, {0xc000417230, 0x1, 0x0?})
	/home/runner/work/ocm/ocm/cmds/ocm/pkg/utils/handling.go:60 +0xeb
[...]

PR #867, opened 2024-08-08
introduce optional repository interface for resolvers

mandelsoft:

CLI command `ocm list componentversions` dumps if only a component is specified together with a resolver (either as part of the config or as option).

This PR fixes this dump.

The resolvers are only able to resolve component versions. Therefore, after the fix the provided list does not show any version for the given component. The expected behaviour would surely be to get the list of versions resolvable by the given resolvers.

Not all resolvers are able to provide repositories offering a version list method. Therefore an optional interface is introduced for
resolvers, which provide potential repositories for a given component. This interface offers the possibility to return repositories or abstract component providers which can be used to get versions or version lists.

This is then checked by the list handling to gather versions provided by all found repositories.

Fixes: #864

### DIFF
--- a/api/ocm/internal/resolver.go
+++ b/api/ocm/internal/resolver.go
@@ -124,6 +124,8 @@ type MatchingResolver struct {
 	rules    []*ResolverRule
 }
 
+var _ ComponentResolver = (*MatchingResolver)(nil)
+
 func NewMatchingResolver(ctx ContextProvider, rules ...*ResolverRule) *MatchingResolver {
 	return &MatchingResolver{
 		lock:  sync.Mutex{},
@@ -169,18 +171,11 @@ func (r *MatchingResolver) LookupComponentVersion(name string, version string) (
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
-	ctx := r.ctx.OCMContext()
 	for _, rule := range r.rules {
 		if rule.Match(name) {
-			repo, cached, err := r.cache.LookupRepository(ctx, rule.spec)
+			repo, err := r.resolveRepository(rule)
 			if err != nil {
 				return nil, err
-			}
-			if !cached {
-				// Even though the matching resolver is closed, there might be components or component versions, which
-				// contain a reference to the repository. Still, it shall be possible to close the matching resolver.
-				refmgmt.Lazy(repo)
-				r.finalize.Close(repo)
 			}
 			cv, err := repo.LookupComponentVersion(name, version)
 			if err == nil && cv != nil {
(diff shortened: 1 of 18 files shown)


api/ocm/types/interface.go
```diff
@@ -9,9 +9,13 @@ type (
 	ContextProvider                  = internal.ContextProvider
 	LocalContextProvider             = internal.LocalContextProvider
 	ComponentVersionResolver         = internal.ComponentVersionResolver
+	ComponentResolver                = internal.ComponentResolver
+	ResolvedComponentProvider        = internal.ResolvedComponentProvider
+	ResolvedomponentVersionProvider  = internal.ResolvedComponentVersionProvider
```
hilmarf, line 14:
```suggestion
	ResolvedComponentVersionProvider = internal.ResolvedComponentVersionProvider
```

api/ocm/resolver.go
```diff
@@ -86,3 +147,5 @@ type MatchingResolver interface {
 func NewMatchingResolver(ctx ContextProvider) MatchingResolver {
 	return internal.NewMatchingResolver(ctx.OCMContext())
 }
+
+type ResolverRule = internal.ResolverRule
```
hilmarf, line 151: Wouldn't it make more sense to put it in here: `api/ocm/interface.go` or `api/ocm/types/interface.go`?

By the way: What's the purpose of 'api/ocm/types/interface.go'? The types are the same as in 'api/ocm/interface.go'! Why do we need this code duplication?

api/ocm/internal/resolver.go
```diff
@@ -18,22 +18,139 @@ import (
...
+// ComponentResolver is an optional interface for a ComponentVersionResolver,
+// which can offer information about potential providers usable to lookup
+// versions for a dedicated component.
+//
+// Be careful, provided objects may refer to closable objects
+// held by the interface provider. They should only be used as long
+// as the object behind the ComponentResolver is not closed.
+type ComponentResolver interface {
+	LookupComponentProviders(name string) []ResolvedComponentProvider
+}
```
hilmarf, line 67: I love the documentation, great! But it's in the `internal` package... :-/ 
As an API user, where would I see this doc then?

api/ocm/types/interface.go
mandelsoft: fixed

api/ocm/resolver.go
mandelsoft: The resolver.go contains some standard resolver implementations. The matching resolver implementation must be in internal, because it is used by the context, therefore it has to be forwarded. To keep all the stuff together, it is in the resolver source file.

api/ocm/internal/resolver.go
mandelsoft: Good question, unfortunately Go does not  forward the documentation to all type aliases, but you can follow the links provided for a type alias.

hilmarf (approved):

PR #880, opened 2024-08-20
fix: typo + missing bin directory

maximiliantech:
This PR fixes typos.
Additionally, it fixes the Make target `make install-requirements` because of missing `bin` directory.

### DIFF
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -65,7 +65,7 @@ ifneq ($(OCI_REGISTRY), $(OCI_REGISTRY_VERSION))
 endif
 
 .PHONY: install-requirements
-install-requirements: $(deps) $(GOPATH)/bin/goimports mkbin mdref
+install-requirements: mkbin $(deps) $(GOPATH)/bin/goimports mdref
 
 .PHONY: mkbin
 mkbin:
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -65,7 +65,11 @@ ifneq ($(OCI_REGISTRY), $(OCI_REGISTRY_VERSION))
 endif
 
 .PHONY: install-requirements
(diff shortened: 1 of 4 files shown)


frewilhelm (approved):

hilmarf (changes requested):
what's the purpose of: `bin/.gitempty` file?
can you please remove it?

hilmarf:
sorry, was too fast with my comment... ok, bin directory is missing... when? Shouldn't the make target ensure that it is created when necessary instead of having an empty file lying around :thought_balloon:

hack/Makefile
```diff
@@ -65,7 +65,11 @@ ifneq ($(OCI_REGISTRY), $(OCI_REGISTRY_VERSION))
 endif
 
 .PHONY: install-requirements
-install-requirements: $(deps) $(GOPATH)/bin/goimports mdref
+install-requirements: $(deps) $(GOPATH)/bin/goimports mkbin mdref
+
+.PHONY: mkbin
```
frewilhelm, line 70: I would expect that the `.PHONY` target is not required. If the target is already present, we don't need the execution of `mkbin`

maximiliantech:
> Shouldn't the make target ensure that it is created when necessary instead of having an empty file lying around 💭

@hilmarf Sure, you are totally right. I was trying out different ways to make the `install-requirements` Make target work. I changed it now so that the `bin` directory will be created if it does not exist.

hack/Makefile
maximiliantech: Fixed with https://github.com/open-component-model/ocm/pull/880/commits/186d7abc99adfbc37c92467f0f07b1e520df614c

hilmarf (approved):

PR #882, opened 2024-08-21
annotate oci layers with component information

fabianburth:
@maximilianbraun a very quick and dirty and in no way production ready experiment to annotate the oci layers with component information. 

Since currently, the abstraction in code is intended to add all kinds of blobs to all kinds of storage backends, a clean solution would require several adjustments. A quick solution could introduce further options containing such metadata. Then the different storage implementation could choose whether to use or ignore those metadata. Furthermore, we might also want more metadata (e.g. to distinct sources from resources and to also annotate the descriptor layer itself).

### DIFF
--- a/api/ocm/extensions/repositories/genericocireg/annotation_test.go
+++ b/api/ocm/extensions/repositories/genericocireg/annotation_test.go
@@ -1,0 +1,86 @@
+package genericocireg_test
+
+import (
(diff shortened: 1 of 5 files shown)


mandelsoft (approved):

mandelsoft:
Maybe, you could also add this spec to the Mapping spec of OCM to OCI Registries.

mandelsoft (approved):

PR #884, opened 2024-08-22
add publich GPG key that can be used for signing

morri-son:
add GPG key to verify signed and uploaded releases


hilmarf (approved):

frewilhelm:
I am not completely happy to put the key into the root-directory of the repository, because it look cluttered.
Can there be any other location? Would it make sense to place it somewhere else?

morri-son:
> I am not completely happy to put the key into the root-directory of the repository, because it looks cluttered. Can there be any other location? Would it make sense to place it somewhere else?

sure, we can even think of putting it into the 

> I am not completely happy to put the key into the root-directory of the repository, because it looks cluttered. Can there be any other location? Would it make sense to place it somewhere else?

We have some 20+ files in the root dir, like Dockerfile, Makefile, central Markdown files, etc. So, does this one really lead to clutter? But for sure, technically we can place it anywhere and reference it.

hilmarf (approved):

hilmarf:
Would it make sense to publish the public.key on the web-site?

hilmarf (approved):

PR #890, opened 2024-08-23
feat: label values as blob

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/13


Skarlso:
The cyclic imports will really be a problem with this one. Because the component version access and all the rest use this package for the identity for example. I will have to shuffle things around to make this work.

Skarlso:
Ok, so I have to shift fetching the blob at the later place. In transit, or during signing.

PR #894, opened 2024-08-26
chore: always create a new release branch and delete all old ones during new release action

hilmarf:

When we create a new release, then let's always be prepared for a potential hotfix and create the release branch.
On the other hand, this will create a lot of unused branches over the time, so the PR will delete all older release branches.
In case of serious situation, where we really need to fix older versions, we still have the tags in place and would need to create manually a patched old release.

releates https://github.com/open-component-model/ocm-project/issues/240


morri-son (approved):
lgtm

PR #899, opened 2024-08-29
release branch creation workflow

mandelsoft:


.github/workflows/release.yaml
```diff
@@ -252,16 +262,6 @@ jobs:
         git push origin ${GITHUB_REF#refs/heads/}
         echo "Next branch version is $v"
 
-    - name: Delete Older Release Branches
```
frewilhelm, line 255: Should we keep all the old release branches? This could become cluttered.

fabianburth (approved):

PR #900, opened 2024-08-30
fix npm input

mandelsoft:

fix problems with npm input handling:
- add missing input registration
- align cli options
- fix typos
- fix html formatting for maven 
- align input/access and blob access (see below)

Accesses for external registries are always handled the same way:
- there is `BlobAccess`implementation for that kind of registry with all the options required (for example credentials, caches,...)
- this `BlobAccess` is then used for the input and access method implementations.

Before, there was only the access method implementation. This PR generalizes this implementation and moves it to
a new `BlobAccess` implementation. The access method implementation has be changed to use this `BlobAccess` type.
The input implementation has been changed to use this implementation, also, by providing all the required options.

The implementations for maven and npm are similar, therefore the option handling for both blob access types
has been aligned.


mandelsoft:
Basically, there is a third application for the BlobAccess, the convenience code bunding the access with the resource/source data to add resources and sources to a CV.

This can always be done by specifying the meta data together with a blob access when adding the artifact.
For some reason there were convenience operations requested bundlings these specification. This required to provide implementations for all combinations of BlobAccess types and resource/source meta data. Those implementations are in package `elements`. Technically this code is not required, we have to decide whether we want to keep it.

If we decide to kept it, such compound implementations have to provided for the NPM BlobAccess, also. I just skipped this, until a final decision has been taken.

Skarlso (dismissed):

Skarlso (approved):

PR #901, opened 2024-08-30
add extended description and remove archlinux

morri-son:


.github/config/goreleaser.yaml
```diff
@@ -97,11 +97,12 @@ nfpms:
...
     homepage: "https://ocm.software/"
     maintainer: "OCM Project <open-component-model@noreply.github.com>"
-    description: "The OCM CLI makes it easy to create component versions and embed them in build processes."
+    description: |
+      "The OCM CLI is used to create and maintain component versions.
+      It can be used to easily embed OCM related actions in CICD processes."
     license: "Apache-2.0"
     formats:
       - deb
-      - archlinux
```
frewilhelm, line 104: why do we remove archlinux?

.github/config/goreleaser.yaml
```diff
@@ -97,11 +97,12 @@ nfpms:
     vendor: "Open Component Model"
     homepage: "https://ocm.software/"
     maintainer: "OCM Project <open-component-model@noreply.github.com>"
-    description: "The OCM CLI makes it easy to create component versions and embed them in build processes."
+    description: |
+      "The OCM CLI is used to create and maintain component versions.
+      It can be used to easily embed OCM related actions in CICD processes."
```
frewilhelm, line 102: If I am not mistaken, then you should not need the `"` on multiline-strings
```suggestion
    description: |
      The OCM CLI is used to create and maintain component versions.
      It can be used to easily embed OCM related actions in CICD processes.
```

.github/config/goreleaser.yaml
morri-son: because the packaging happens in another action that Hilmar just works on and that you commented on already :-) I'm about to move out Debian, too.

frewilhelm (approved):
lgtm - but we should wait for a review from @hilmarf

PR #903, opened 2024-08-30
chore: nightly build

hilmarf:

### DIFF
--- a/.github/config/goreleaser.yml
+++ b/.github/config/goreleaser.yml
@@ -1,0 +1,111 @@
+version: 2
+before:
+  hooks:
+    - go mod tidy
+    - make CTF_TYPE=directory ctf
+
+builds:
+  - <<: &build_defaults
+      binary: ocm
+      main: ./cmds/ocm/main.go
+      ldflags:
+        - -s -w -X ocm.software/ocm/api/version.gitVersion={{.Version}} -X ocm.software/ocm/api/version.gitCommit={{.Commit}} -X ocm.software/ocm/api/version.buildDate={{.CommitDate}}
+      env:
+        - CGO_ENABLED=0
+    id: linux
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+      - arm
+    goarm:
+      - 7
+  - <<: *build_defaults
+    id: darwin
+    goos:
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+  - <<: *build_defaults
+    id: windows
+    goos:
+      - windows
+archives:
+  - name_template: "{{ .Binary }}-{{ .Version }}-{{ .Os }}-{{ .Arch }}"
+    id: nix
+    builds: [linux, darwin]
+    format: tar.gz
+    files:
+      - none*
+  - name_template: "{{ .Binary }}-{{ .Version }}-{{ .Os }}-{{ .Arch }}"
+    id: windows
+    builds: [windows]
+    format: zip
+    files:
+      - none*
+  - name_template: "{{ .ProjectName }}-{{ .Version }}-ctf"
+    format: tgz
+    meta: true
+    files:
+    - src: gen/ctf
+      dst: .
+
+source:
+  enabled: true
+  name_template: '{{ .ProjectName }}-{{ .Version }}-source_code'
+sboms:
+  - id: source
+    artifacts: source
+    documents:
+      - "{{ .ProjectName }}-{{ .Version }}-sbom.spdx.json"
+signs:
+  - cmd: cosign
+    env:
+      - COSIGN_EXPERIMENTAL=1
+    certificate: '${artifact}.pem'
+    args:
+      - sign-blob
+      - '--output-certificate=${certificate}'
+      - '--output-signature=${signature}'
+      - '${artifact}'
+      - '--yes'
+    artifacts: checksum
+    output: true
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
+brews:
+  - name: ocm
+    repository:
+      owner: open-component-model
+      name: homebrew-tap
+      token: "{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}"
+    directory: Formula
+    homepage: "https://ocm.software/"
+    description: "The OCM CLI makes it easy to create component versions and embed them in build processes."
+    test: |
+      system "#{bin}/ocm --version"
+nfpms:
+  - id: debian
+    package_name: ocm-cli
+    builds: [linux]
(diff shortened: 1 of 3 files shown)


mandelsoft:
Why do you want the goreleaser every night?

hilmarf:
> Why do you want the goreleaser every night?

see: https://github.com/open-component-model/ocm-project/issues/240

> Every night there should be a build of the latest sources from the main branch. The used commit should be tagged with "nightly" and there should be one (pre-created) release also called "nightly". That nightly release should get assigned with the new nightly tag on every successful build. The assets of the release for sure should be the ones created by the last nightly build. Instead of creating a new Docker image with the OCM CLI on every push to the main branch, the nightly build should push that image and tag it with "latest".

mandelsoft:
But they you have to change the prerelease version part somehow to `nightly` instead of `dev` for all the involved image and component versions. The initial idea was just to use `dev` for such floating intermediate builds/releases.
You just call the components workflow, which will produce the dev version, I think.

morri-son:
@hilmarf 
> You just call the components workflow, which will produce the dev version, I think.
will injecting the nightly version be sufficient to get persisted everywhere, or is there a kind of hardcoded ref to "dev" based on the prerelease?

frewilhelm:
we could talk about the trigger for the build not happening on every change, but only when certain files (e.g. *.go) are changed (or vice versa, the build should not start when certain files are changed, e.g. *.md).
However, the discussion could become very detailed and the maintenance effort could be quite high (you always have to keep this in mind when adapting the technical release process).

What do you think?

.github/workflows/push_ocm.yaml
```diff
@@ -70,3 +62,39 @@ jobs:
...
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: '${{ github.workspace }}/go.mod'
+        check-latest: false
+    - name: Run GoReleaser --nightly
+      uses: goreleaser/goreleaser-action@v6
+      with:
+        distribution: goreleaser
+        version: 2.2.0
+        args: release --clean --nightly --timeout 60m --skip=validate --config=.github/config/latest.yml
```
frewilhelm, line 97: Looking at the [references](https://goreleaser.com/customization/nightlies/) it is described as `goreleaser` pro function. Do we have a pro license?

hilmarf: damn it... I thought, last time, when I checked, it wasn't a PRO feature...

morri-son: costs 50 Euro per month per Enterprise. We may check if somebody at SAP already has such a license...

hilmarf: found another solution... please review again and spend me the 50 bucks ;-)

morri-son (commented):
would we like to set v0.0.0 or use the vx.y.0-dev?

hilmarf:
> would we like to set v0.0.0 or use the vx.y.0-dev?

1. version should be semver-compatible
2. when we use one single stable version number, it's a lot easier for all self-made automation scripts of our consumers, to fetch the latest version - that's why I started with v0.0.0

.github/workflows/push_ocm.yaml
```diff
@@ -70,3 +67,67 @@ jobs:
...
+      with:
+        fetch-depth: 0
+        token: ${{ steps.generate_token.outputs.token }}
+    - name: Enable gh-CLI
+      uses: actions4gh/setup-gh@v1.0.2
+      with:
+        token: ${{ steps.generate_token.outputs.token }}
+    - name: delete release and move tag
+      run: |
+        if [ "$(gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /repos/open-component-model/ocm/releases | grep -c 'ocm-cli/latest')" -ge 1 ]; then
```
frewilhelm, line 98: Where does this date `2022-11-28` come from? Do we have to update this in the future?

.github/workflows/push_ocm.yaml
```diff
@@ -70,3 +67,67 @@ jobs:
...
+        version: 2.2.0
+        args: release --clean --timeout 60m --snapshot --skip=validate --config=.github/config/latest.yml
+      env:
+        GITHUBORG: ${{ github.repository_owner }}
+        GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+        GORELEASER_CURRENT_TAG: v0.0.0
+        GORELEASER_PREVIOUS_TAG: v0.0.0
+    - name: remove some meta files
+      run: |
+        rm -rf dist/*.json dist/*.yaml
```
frewilhelm, line 121: Why do we need the `--force` flag? (i am always anxious seeing `rm -f` in any script^^)

.github/workflows/push_ocm.yaml
```diff
@@ -70,3 +67,67 @@ jobs:
...
+    - name: Create Release
+      uses: Hs1r1us/Release-AIO@v2.0.0
```
frewilhelm, line 123: This action does not seem to be maintained actively. The last commit was two years ago and there is still a TODO open (see their README.md).

Maybe this action could be more suitable? https://github.com/ncipollo/release-action/tree/main

I did not look too deep into it, but it seems that it also provides a functionality to make `latest` releases and replace artifacts.

.github/workflows/push_ocm.yaml
frewilhelm: I guess this is the Api-Version for GitHub. However, this seems an odd identifier for the API imo.^^

hilmarf: sorry, but it's regular github stuff - https://docs.github.com/de/rest/about-the-rest-api/api-versions

.github/workflows/push_ocm.yaml
```diff
@@ -5,59 +5,56 @@ on:
   push:
     branches:
       - main
+    paths:
+      - '**/*.go'
+      - '**/*.mod'
+      - '**/*.sum'
```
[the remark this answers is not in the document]
hilmarf: nope, that's correct - currently we have only one

but what's the point when we ensure, we're prepared already for any upcoming further restructuring? e.g. splitting the cli from the library

.github/workflows/push_ocm.yaml
hilmarf: not necessary anymore with the other GHA - thanks for the hint!

.github/workflows/push_ocm.yaml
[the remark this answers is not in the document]
frewilhelm: True. Additionally, it does not make any difference if it says `**/*.mod` or `go.mod`. Fine for me!

PR #904, opened 2024-08-30
feat: replace docker with oras

Skarlso:
This is a highly Draft pull request understanding the nuances of the OCI package.

There are a LOT of things to do yet:
- [x] Figure out to nicely create repositories with hosts
- [x] Add authentication configuration to the OCI repository creation
- [x] Figure out which store to use and which implementation to use where
- [x] Figure out the least intrusive modification that still works
  - this will partially be done via declared interfaces to support previous behaviour


Skarlso:
Okay, I made the thing work, but it's kinda an ugly workaround I don't much like. But we might have to live with. Let's see how it play out.

```
REGISTRY REPOSITORY                       KIND     TAG     DIGEST
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.1.2  sha256:0d88d0ee812e10ef2683956bbb6a3d13ff254fd5a80395a61850969672bc458a
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.6  sha256:1192b8e7b091a50d721a588cc8952bce28d8a903441399b24109da8106f9447f
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.4  sha256:1acbd43cec6e30007556af51057b4f9acb6266932f805350e9971f188f5dda54
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.4.1  sha256:1c99548424582f7c01bfe49f626ca0d4a89ba88723c78b782f922beff3451a71
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.1.1  sha256:3461e5031ebb93f97c4a0decd58bcc947f33ad7f7ef82ce30d8ffdc3540be59b
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.6.0  sha256:3695b88bae23f70f2b4378dd84e2d52ebb7abd2559a095f1f7dc813a6d8a63d5
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.4.2  sha256:3be7403dbe91031654ce30c4528babcc48c1b72e9697fadca2a0f3db73bf1c34
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.5.3  sha256:3f2a56b16b9167ed338a757cf079c78191ef1f4ceab83860b2d486bdf6a2c0ad
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.7.0  sha256:428efd9683e63365341601f3e4c647afe39433093ba0dec00edd83bb16aefe5b
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.5  sha256:494d41c43ed7ef1e9288faa61eb16cac42f282d1dc82ee292e9fc9e8509f6041
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.2  sha256:5018d246d97cd139990c731b307e55592805ca08296db17fe5742ca6d552d973
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.5.1  sha256:5d2d6273541ea6639e709333a9d1d2b9bcf14103de9005d7d5dedfb1002b71fc
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.5.0  sha256:5dda48e405509881bb709bd0f3dc071e39951f97145d90a6da3850e12720cd79
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.2.0  sha256:6bcdf8db8b903a852a7360078107cea686fbc8539b180a3a7d677408b5e3fb4a
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.11 sha256:734843cf22fe96add178b5c066dda49677145e5f8e13c7161054e67f21883afa
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.10 sha256:7f16a7d6bc790060e99c7973a5ac8ab2218221235497e4357933f6585c284787
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.1.0  sha256:80660ce2b58a7f6ee1429478009bcd4710a8c07d73da312b89f784a171e0f07a
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.8.0  sha256:82d91fbe73d72ac8001cb43ec7dcc371428bad6960a1d523a6c96f752431281c
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.0.1  sha256:8a20a73c484dfaea11328384d3bd00410778027144144d4d32a21721ef8d12a4
ghcr.io  skarlso/crd-bootstrap-controller manifest latest  sha256:c78d260d00a7a6dfc4e6174fd98bbe0445a87272cd15c7b8ac863befe40b17cc
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.9.0  sha256:c78d260d00a7a6dfc4e6174fd98bbe0445a87272cd15c7b8ac863befe40b17cc
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.7  sha256:c8ff7a9b501c97489f15ccfee70a1d34544ad60e17e475964a68ce8d65102d8e
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.3.9  sha256:d4ec1b234fd4de93f3de9c007f6602399fdea12816a6698beb10019c7d90523a
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.4.0  sha256:d9ec40eb5d3c74bb1bf94b5981d20170517f2adfcf82cef9511074c9be2dd9a9
ghcr.io  skarlso/crd-bootstrap-controller manifest v0.5.2  sha256:ea2f4ac283ddb91484530550f52aa20cd11c8fe20e6a2dfc9f58710684ccb9d4
```

Right now, there is a "problem" with ORAS where you can't fetch anything without first knowing the Size. We can gather the size, by doing a Resolve call beforehand. But that needed to include the REF where there is a GetBlobData call using OCI. That needed a bit of refactoring passing it around.

Now, docker also did this but it did it separately and included in the client. I might also make the same refactor here. But suffice to say, the Ref needs to be passed around for oras to work right now.

Now, doing a Resolve should be fine even if delay is set, because it's not a Fetch call. So no data is retrieved. But it does retrieve the size. So doing something like this works:
resolvedDescriptor, err := resolver.Resolve(dummyContext, ref+"@"+digest.String())

But it's kinda a hack and I don't necessarily like it. Let's see what oras peeps reply to my issue on fetching without size because removing their size check makes everything work supernicely. So I don't really understand why they would have that restriction ¯\_(ツ)_/¯.

Let's see how it goes.

The issue https://github.com/oras-project/oras-go/issues/822

Skarlso:
This is odd. Looks like linter is failing locally, but doesn't fail in Github. Very interesting.

hilmarf:
> This is odd. Looks like linter is failing locally, but doesn't fail in Github. Very interesting.

different linter version? different call parameters?

Skarlso:
I'm pretty sure the linter isn't running actually.

Skarlso:
Yeah I think I know what's up.

Skarlso:
Ah, it won't run the new linter until it's merged. That's a bummer.

Skarlso:
I fixed the linter issues. Once this is merged, it should be okay. :)

Makefile
```diff
@@ -62,8 +68,8 @@ format:
 	@$(REPO_ROOT)/hack/format.sh $(EFFECTIVE_DIRECTORIES)
 
 .PHONY: check
-check:
-	@$(REPO_ROOT)/hack/check.sh --golangci-lint-config=./.golangci.yaml $(EFFECTIVE_DIRECTORIES)
```
hilmarf, line 66: where/which is now the configuration been used for golangci-lint?

Skarlso: It's the configuration at the root by default called `.golangci.yaml`.

hack/check.sh
```diff
@@ -1,28 +0,0 @@
...
-    shift
-    ;;
-  esac
-done
-
-echo "> Check"
-
-echo "Executing golangci-lint"
-echo "  golangci-lint run $GOLANGCI_LINT_CONFIG_FILE $opt --timeout 10m $@"
-golangci-lint run $GOLANGCI_LINT_CONFIG_FILE $opt --timeout 10m $@
```
hilmarf, line 26: I don't know, why there was a timeout introduced in the first place... but now it's not set. Are we sure, it's ok to run without it?

Skarlso: YOu're right, I definitely need to put that timeout back. It's now 5m now.

Makefile
hilmarf: hmm... ok... @frewilhelm what do you think, shall we move that config also under: `.github/config`? Since it's used by `make check` which runs `golangci-lint` I'm not sure about it.

frewilhelm: Hm, I would move `.golangci.yaml` to `.github/config` because (1) the root-repository should look less cluttered, (2) we moved all other config-files there, (3) and `golangci-lint` is primarily used in our GitHub-Actions. One can use it locally, but that is not the main purpose.

Skarlso: Sure, I can move the config there. No problem.

Skarlso: Done. Moved the config.

morri-son:
@Skarlso do we only for @fabianburth and/or @mandelsoft to review this before merging or are there remaining tasks?

Skarlso:
Yes we are waiting for Uwe.

api/oci/extensions/repositories/ocireg/namespace.go
```diff
@@ -69,26 +53,25 @@ func (n *NamespaceContainer) SetImplementation(impl support.NamespaceAccessImpl)
 	n.impl = impl
 }
 
-func (n *NamespaceContainer) getPusher(vers string) (resolve.Pusher, error) {
+func (n *NamespaceContainer) getPusher(vers string) (content.Pusher, error) {
```
mandelsoft, line 56: `getPusher` just provides a remote.Repository object  with identical initial attributes as the one stored in n.ociRepo, but with many complex operations repeated for every push.

Maybe this could be simplified by directly creating it with the `Reference` and `Client` set from the `ociRepo`. Only the tag/digest has to be set to `pusher.Reference.Reference`.

api/oci/extensions/repositories/ocireg/namespace.go
```diff
@@ -159,18 +151,18 @@ func (n *NamespaceContainer) GetArtifact(i support.NamespaceAccessImpl, vers str
 	}
 	_, acc, err := blobData.GetBlobData(desc.Digest)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("get blob data failed for digest %s: %w", desc.Digest, err)
```
mandelsoft, line 154: here you know, that a manifest is fetched, this should be stated in the message if you create a detailed message.

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -112,7 +115,7 @@ func (r *RepositoryImpl) getCreds(comp string) (credentials.Credentials, error)
 	return identity.GetCredentials(r.GetContext(), r.info.Locator, comp)
 }
 
-func (r *RepositoryImpl) getResolver(comp string) (resolve.Resolver, error) {
+func (r *RepositoryImpl) getResolver(ref string, comp string) (registry.Repository, error) {
 	creds, err := r.getCreds(comp)
 	if err != nil {
 		if !errors.IsErrUnknownKind(err, credentials.KIND_CONSUMER) {
```
mandelsoft, line 121: see comment above, instead, provide a second method used to prepare a pusher for a version push based on the .ociRepo attribute already resolved.

api/ocm/tools/signing/digest.go
```diff
@@ -67,13 +67,13 @@ func VerifyResourceDigest(cv ocm.ComponentVersionAccess, i int, bacc ocm.DataAcc
 	hasher := registry.GetHasher(dtype.HashAlgorithm)
 	digest, err := octx.BlobDigesters().DetermineDigests(raw.Type, hasher, registry, meth, req...)
 	if err != nil {
-		return false, errors.Wrapf(err, resMsg(raw, acc.Describe(octx), "failed determining digest for resource"))
+		return false, errors.Wrap(err, resMsg(raw, acc.Describe(octx), "failed determining digest for resource"))
```
mandelsoft, line 70: yes, you are right! Good point...

api/version/version.go
```diff
@@ -55,24 +55,24 @@ func Current() string {
...
 	)
 
 	v, err := semver.NewVersion(gitVersion)
 	if err == nil {
-		gitMajor = strconv.Itoa(int(v.Major()))
-		gitMinor = strconv.Itoa(int(v.Minor()))
-		gitPatch = strconv.Itoa(int(v.Patch()))
+		gitMajor = strconv.FormatUint(v.Major(), 10)
+		gitMinor = strconv.FormatUint(v.Minor(), 10)
+		gitPatch = strconv.FormatUint(v.Patch(), 10)
```
mandelsoft, line 75: Ok, but you have to specify more irrelevant info (10), so why not directly using the method describing what you want to do.

cmds/ocm/common/output/table.go
```diff
@@ -9,8 +9,8 @@ import (
 
 func FormatTable(ctx Context, gap string, data [][]string) {
 	columns := []int{}
-	max := 0
-	maxtitle := 0
+	maxValue := 0
+	maxTitle := 0
```
mandelsoft, line 14: Is all this a linter problem, or why have you touched so many files not related to the PR. By he way, the linter version specified in the requirements does NOT complain about this. Why did you use another one.

But anyway, in this case here, the change is really useful, because there are different max values. But all the other locations.....

docs/reference/ocm_add_componentversions.md
```diff
@@ -160,28 +160,6 @@ are configured for the operation. It has the following format
 </center>
 
 The uploader name may be a path expression with the following possibilities:
-  - <code>ocm/npmPackage</code>: uploading npm artifacts
-
-    The <code>ocm/npmPackage</code> uploader is able to upload npm artifacts
-    as artifact archive according to the npm package spec.
```
mandelsoft, line 166: Hmm, why has the order changed?  may be there is a sort missing for the generator to achieve a stable order.

yes, in cmds/ocm/commands/common/options/uploaderoption/option.go (Usage) the list should be sorted by the Name attribute.
Similar problem in downloaderoption.

mandelsoft (changes requested):

api/credentials/utils.go
```diff
@@ -109,25 +109,25 @@ func GuessConsumerType(ctxp ContextProvider, spec string) string {
 			}
 		}
 		if fix == "" {
-			min := -1
+			minVal := -1
```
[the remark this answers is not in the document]
Skarlso: Because `min` is a defined function in Go 1.23 onwards.

api/oci/cpi/support/flavor_index.go
```diff
@@ -94,6 +94,7 @@ func (i *IndexAccess) GetBlob(digest digest.Digest) (internal.BlobAccess, error)
 		}
 		return blobaccess.ForDataAccess(d.Digest, d.Size, d.MediaType, data), nil
 	}
+
```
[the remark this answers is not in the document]
Skarlso: Most likely automatic fixes? I'll remove them.

api/oci/extensions/repositories/ocireg/blobs.go
```diff
@@ -95,13 +89,17 @@ func NewBlobContainer(cache accessio.BlobCache, mime string, fetcher resolve.Fet
 }
 
 func (n *blobContainer) GetBlobData(digest digest.Digest) (int64, cpi.DataAccess, error) {
-	logrus.Debugf("orig get %s %s\n", n.mime, digest)
-	acc, err := NewDataAccess(n.fetcher, digest, n.mime, false)
-	return blobaccess.BLOB_UNKNOWN_SIZE, acc, err
+	logrus.Debugf("orig get %s\n", digest)
```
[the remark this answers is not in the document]
Skarlso: Good idea!

api/oci/extensions/repositories/ocireg/blobs.go
```diff
@@ -23,25 +23,21 @@ type BlobContainer interface {
...
-	pusher  resolve.Pusher
-	mimes   map[string]BlobContainer
+	lock  sync.Mutex
+	cache accessio.BlobCache
+	repo  registry.Repository
+	mimes map[string]BlobContainer
 }
 
-func NewBlobContainers(ctx cpi.Context, fetcher remotes.Fetcher, pusher resolve.Pusher) *BlobContainers {
+func NewBlobContainers(ctx cpi.Context, repo registry.Repository) *BlobContainers {
```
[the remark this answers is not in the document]
Skarlso: Nice, sounds good to me. I wasn't completely sure which parts I could further remove here.

api/version/version.go
Skarlso: It was warning about overflow casting because it was casting uint64 into and int. And FormaUint is type safe with the 10 at the end. I'll create a constant for the ten.

cmds/ocm/common/output/table.go
Skarlso: Because `max` and `min` are now reserved functions. :)

docs/reference/ocm_add_componentversions.md
Skarlso: I'll add the sort 👍

mandelsoft: I'm already doing some fixes, should I add it?

Makefile
```diff
@@ -9,6 +9,12 @@ COMMIT                                         := $(shell git rev-parse --verify
 
 CONTROLLER_TOOLS_VERSION ?= v0.14.0
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+GOLANGCI_LINT ?= $(LOCALBIN)/golangci-lint
+
+.PHONY: golangci-lint
+golangci-lint: $(GOLANGCI_LINT)
+$(GOLANGCI_LINT): $(LOCALBIN)
```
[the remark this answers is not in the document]
Skarlso: Argh, I could have sworn I added that.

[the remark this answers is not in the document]
Skarlso: ah okay, I'll use that one, and remove this here.

docs/reference/ocm_add_componentversions.md
Skarlso: Yes please, thank you.

api/oci/extensions/repositories/ocireg/namespace.go
Skarlso: Not actually sure what you mean here. :) GetPusher just returns the ociRepo as it is. It doesn't do anything other with it but to make sure that it is set and configured. If we don't have a digest, we configure the pusher here with authentication and everything.

What Reference are you referring to? There is no `Reference` or `Client` for the ociRepo. 🤔

Makefile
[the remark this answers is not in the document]
Skarlso: done

cmds/ocm/common/output/table.go
Skarlso: Yah, that was an interesting choice... could have been in math or something.

api/oci/extensions/repositories/ocireg/namespace.go
mandelsoft: The type is `registry.Repository`which is an interface, you are right, but it is filled
by the `getResolver()` method, which creates a `remote.NewRepository(ref)` object, which is 
a struct with the fields mentioned above.

Skarlso: Are you suggesting to only construct a `&remote.Repositroy{}` with the Client and Reference from `getResolver`?

I don't really think there is a benefit to that. 🤔

This appears to be a cleaner approach:

```go
func (n *NamespaceContainer) getPusher(tag string) (content.Pusher, error) {
	if ok, _ := artdesc.IsDigest(tag); ok {
		return n.ociRepo, nil
	}

	ref := n.repo.GetRef(n.impl.GetNamespace(), tag)
	n.repo.GetContext().Logger().Trace("get pusher", "ref", ref)

	repo, err := n.repo.getResolver(ref, n.impl.GetNamespace())
	if err != nil {
		return nil, fmt.Errorf("unable get resolver: %w", err)
	}

	return repo, nil
}
```

I don't like returning an interface, but in this case, it's cleaner. Further down the line there are cases where tag is passed in or not that decides what pusher to call. So eliminating this function is actually not trivial unless I refactor the whole thing.

Skarlso: I completely removed this and replaced it with PushReference. It seems to be working fine.

```
ocm oci artifacts get ghcr.io/skarlso/component-descriptors/github.com/acme/podinfo
REGISTRY REPOSITORY                                            KIND     TAG    DIGEST
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v1.0.0 sha256:1d0d559190a542f37baf49df6176eacbcc209dd79aade1b558a14d47e19fcf20
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v2.0.0 sha256:262cdb76d74ebc27e91281024d8e39b975c290b82dd5ace19a78fcfe6efb99f7
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v1.0.5 sha256:32e7197eb2e5e6bb14836cfa4f0b0ba6dd8208f7fd1d928179267f7c873b77c3
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v5.0.1 sha256:66044b21749d0c674af6b114c9630395e008db2acf51fbca7d9c5b3ca07c7582
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v3.0.0 sha256:962f9a8f3f0dc115e8fbe50bca9c518c3b75dd82ebaa60b7af64c0c71b80feba
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v6.0.0 sha256:9ca8474f40ec9e68fed82ce0e670beacf17565defecbbd00aefb9c171f37fafb
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v5.0.0 sha256:9e03d002b723dd14d368b5ba232f05f38384edb071f37f859ce03475b11bc868
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v2.0.3 sha256:b489776734b2060fda5c5c4f8f0a1874204587648efef1bd91dce2c625123504
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v2.0.4 sha256:d897331fb87a3776842e2ad4264d77b06b8d125b21c4a1f8152574e3b5fffd2c
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v4.0.0 sha256:defdf16282e2456f940a426ef01bfc0d326832395361deae498e78389a384675
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v4.0.2 sha256:f22bd423ca0b0bf04796c73634a71ed7e056959308256aea1182f4626a9662e0
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v4.0.1 sha256:f7c7cacb667da08bddd89478982858d83aaef539ff79e60f3a45efde400bebb5
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v2.0.1 sha256:f9aa0286e0395f7b1638718f0b7d07f9c8c4400dc3843814e7a93019e27ef634
ghcr.io  skarlso/component-descriptors/github.com/acme/podinfo manifest v2.0.2 sha256:fb41aeb216afbce8d91ff5b0f12ac00d7af6122daaacfa75071b1d279d812b21

```

Was able to get private repo data as well.

Skarlso: And transfer and ocm get also worked.

api/oci/extensions/repositories/ocireg/namespace.go
```diff
@@ -208,43 +163,54 @@ func (n *NamespaceContainer) AddArtifact(artifact cpi.Artifact, tags ...string)
...
 	return blob, err
 }
 
+func (n *NamespaceContainer) pushTag(blob blobaccess.BlobAccess, tag string) error {
+	reader, err := blob.Reader()
+	if err != nil {
+		return err
+	}
+	expectedDescriptor := *artdesc.DefaultBlobDescriptor(blob)
+	if err := n.ociRepo.PushReference(context.Background(), expectedDescriptor, reader, tag); err != nil {
```
mandelsoft, line 188: Ah, there is an explicit ref pusher interface, even better.,

mandelsoft (approved):

PR #906, opened 2024-09-02
align package names with folder names

mandelsoft:
Because of an error in the migration script some package renames resulted in a mismatch of the folder name
of the package and its name.

This PR fixes this, and also aligns some packages which had the wrong names even before the migration.

### DIFF
--- a/api/utils/misc/printer_test.go
+++ b/api/utils/misc/printer_test.go
@@ -1,4 +1,4 @@
-package common
+package misc
 
 import (
 	"bytes"
--- a/api/cli/builder.go
+++ b/api/cli/builder.go
@@ -1,4 +1,4 @@
-package clictx
+package cli
 
 import (
 	"context"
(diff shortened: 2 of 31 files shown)


api/utils/misc/printer_test.go
```diff
@@ -1,4 +1,4 @@
-package common
+package misc
```
frewilhelm, line 1: Shouldn't this be `misc_test` like in `api/utils/misc/history_test.go`?

mandelsoft: It wasn't `_test` before to be able to use the package without import.

fabianburth (approved):

api/utils/misc/printer_test.go
frewilhelm: Why would somebody want to import these tests?

PR #915, opened 2024-09-10
feat: remove deprecated functions

Skarlso:


api/ocm/selectors/json.go
```diff
@@ -223,3 +224,14 @@ func (j JSONSchemaSelector) Match(obj map[string]string) (bool, error) {
 	}
 	return res.Valid(), nil
 }
+
+// TODO 1: this might not work.
+// TODO 2: what about accessor2?
```
hilmarf, line 229: are you done with this already? The rest LGTM

Skarlso: No, I need confirmation that this makes sense there. I have no idea how this works. :D

Skarlso: Nice, this point is now "fixed". :D

Skarlso:
Note that the Docker section still has deprecated stuff, but we aren't removing that here, because the entire thing is removed in #904

hilmarf (approved):
LGTM

PR #917, opened 2024-09-11
fix: runner caching code that it uses

Skarlso:

### DIFF
--- a/.github/workflows/lint_and_test.yaml
+++ b/.github/workflows/lint_and_test.yaml
@@ -1,11 +1,7 @@
 name: lint-and-test
 
 on:
-  pull_request_target:
-  push:
-    branches:
-      - main
-  workflow_dispatch:
+  pull_request:
   workflow_call:
 
 permissions:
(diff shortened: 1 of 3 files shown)


Skarlso:
```
ok      ocm.software/ocm/api/utils/template     12.614s
FAIL
make: *** [test] Error 1

```

It should have failed 😢

hilmarf (approved):

Skarlso:
Alright, let's hope that it fails now...

.github/workflows/lint_and_test.yaml
```diff
@@ -1,11 +1,7 @@
 name: lint-and-test
 
 on:
-  pull_request_target:
-  push:
-    branches:
-      - main
-  workflow_dispatch:
+  pull_request:
   workflow_call:
```
hilmarf, line 5: which one of these is actually required, when we want to run this during release?
https://github.com/open-component-model/ocm/blob/main/.github/workflows/release.yaml#L101

Skarlso: `workflow_call` I believe.

hilmarf (approved):

PR #923, opened 2024-09-13
feat: automatically label all issues with ipcei

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/248


.github/workflows/auto-label-issues.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - opened
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - run: gh issue edit "$NUMBER" --add-label "$LABELS"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hilmarf, line 14: Is this still a valid approach? The other actions normally use this bot-app to retrieve a token...

hilmarf (approved):

.github/workflows/auto-label-issues.yaml
Skarlso: The plain github token is enough for this process with the right permission. There is no need for the bot account level token.

PR #927, opened 2024-09-16
Custom Unmarshaler for Consumer Identity

fabianburth:
This PR adds a custom unmarshaler to the consumer identity type.

Adding this custom unmarshaler to the consumer identity allows a yaml specification containing a data type other string, e.g. a hostpath spec with a port. Previously, it would error if the user specified `port: 5000` and instead, the user had to specify `port: "5000"`.

### DIFF
--- a/api/credentials/internal/identity.go
+++ b/api/credentials/internal/identity.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 
 	"github.com/sirupsen/logrus"
+
+	"ocm.software/ocm/api/utils/runtime"
 )
 
 func init() {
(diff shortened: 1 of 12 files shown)


frewilhelm (approved):

ikhandamirov (approved):

api/credentials/internal/identity.go
```diff
@@ -91,6 +93,26 @@ func orMatcher(list []IdentityMatcher) IdentityMatcher {
 // ConsumerIdentity describes the identity of a credential consumer.
 type ConsumerIdentity map[string]string
 
+func (c *ConsumerIdentity) UnmarshalJSON(data []byte) error {
+	m := map[string]interface{}{}
+	err := runtime.DefaultJSONEncoding.Unmarshal(data, &m)
+	if err != nil {
```
mandelsoft, line 99: should be 
```
var m  map[string]interface{}
```
without initialization

mandelsoft (changes requested):

api/credentials/internal/identity.go
fabianburth: why?

mandelsoft (approved):

api/credentials/internal/identity.go
mandelsoft: Depends on the implementation. For example your unmarshaller always creates a new object. I would not rely on this, to get a clean result,

PR #929, opened 2024-09-16
fix busy text file during plugin installation

mandelsoft:

A binary just written cannot be executed while the file is still open for write (even on Unix).
Therefore, we have to close the file for the plugin executable, before it can be executed to extract
the plugin descriptor.


fabianburth (approved):
A short description of the issue and the proposed solution in the pull request description would be great!

PR #934, opened 2024-09-19
some fixes and alignments

mandelsoft:
- apply sorting to output of handler descriptions for up- and downloaders
- improve tree output element title handling


Skarlso (approved):
nice fix! :)

PR #937, opened 2024-09-23
feat: Extend plugin API to support transfer handlers

mandelsoft:

The OCM library already provides an extension point to provide transfer handlers
controlling the transport of component versions from one environment into another one.

A transfer handler controls:
- how to handle changes in component versions when re-transporting them 
- how to resolve component references and whether they should be transported, also
- whether sources and resources should be transported by value or by reference.

So far, there were two implementations:
- a standard implementation working on the standard transfer options, like `--recursive` or `--copy-resources`
- a *spiff++* based implementation based on the standard handler adding the capability to influence
  the behaviour based on a spiff script.

On the command line interface always the spiff handler was used which works like the standard handler if no script
is configured.

This PR adds three new features:
- it supports a hierarchical naming scheme to denote transfer handlers (like for download and upload handlers),
  which can be used un the command line to select and configure supported transfer handler types.
- it extends the *Plugin* API to support transfer handler. This way a plugin may implement the transfer handler interface to
  provide an arbitrary rule set controlling the transport process.
- A general configuration option for handlers allows for passing additional handler configuration

Similar to the uploaders and downloaders the naming scheme for transfer handlers looks as follows:
- `ocm/<builtin handlers>`:  used for the already existing  `standard` and `spiff` handlers.
  Additional handlers may now add themselves to appropriate names to be usable from the command line.
- `plugin/'` A dedicated handler maps plugin provided transfer handlers into the naming scheme.
  It uses the sub scheme `<plugin name>[/<handler name>]`. A plugin may provide multiple (named) handlers.


api/ocm/plugin/ppi/utils.go
```diff
@@ -115,6 +119,79 @@ func (b *nameDescription) Description() string {
...
+type testDH struct {
+	DecisionHandlerBase
+}
+
+func (d *testDH) DecideOn(p Plugin, question interface{}) (bool, error) {
+	panic("just a type test")
+}
+
+var _ DecisionHandler = (*testDH)(nil)
+
```
fabianburth, line 192:
```suggestion
```

remove

fabianburth (changes requested):

api/ocm/plugin/ppi/interface.go
```diff
@@ -211,3 +219,28 @@ type Command interface {
...
+	GetQuestions() []DecisionHandler
+}
+
+type DecisionHandler interface {
+	GetQuestion() string
+	GetDescription() string
+	GetLabels() *[]string
+
+	DecideOn(p Plugin, question interface{}) (bool, error)
+}
```
fabianburth, line 235: Considering these are the interfaces developers are supposed to implement, I think they could use a little more documentation.

cmds/transferplugin/transferhandlers/demo_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"ocm.software/ocm/api/ocm/plugin/common"
+	"ocm.software/ocm/api/ocm/plugin/ppi"
+	"ocm.software/ocm/api/ocm/plugin/ppi/cmds"
+	"ocm.software/ocm/api/utils"
+	"ocm.software/ocm/api/utils/misc"
+	"ocm.software/ocm/cmds/transferplugin/app"
+)
+
+var _ = Describe("Test Environment", func() {
+	It("", func() {
```
fabianburth, line 26: test should have a description

api/ocm/plugin/descriptor/descriptor.go
```diff
@@ -235,6 +239,33 @@ func (a CommandDescriptor) GetDescription() string {
...
+func (d TransferHandlerDescriptor) GetQuestion(name string) *QuestionDescriptor {
+	for _, q := range d.Questions {
+		if q.Question == name {
+			return &q
+		}
+	}
+	return nil
+}
+
+type QuestionDescriptor struct {
```
fabianburth, line 261: needs more documentation (especially, what is the labels property intended for)

api/ocm/plugin/internal/transfer.go
```diff
@@ -0,0 +1,91 @@
...
+
+// Resolution describes the transport context for a component
+// version, including the new handler specification and
+// the source repository to use to look up the reference.
+type Resolution struct {
+	RepositorySpec *ocm.GenericRepositorySpec `json:"repository,omitempty"`
+	// TransferHandler is the handler identity according to the transfer handler
+	// name scheme.
+	TransferHandler string           `json:"transferHandler,omitempty"`
+	TransferOptions *TransferOptions `json:"transferOptions,omitempty"`
```
fabianburth, line 56: add documentation why the transfer handlers returns the transfer options (= because it might change the options).

api/ocm/plugin/ppi/interface.go
```diff
@@ -211,3 +219,49 @@ type Command interface {
...
+
+	GetDescription() string
+	// GetLabels returns the list of labels, which should be passed
+	// to the transfer handler. If nothing is specified all labels
+	// are transferred, if an empty list is given no label is handed over
+	// to the plugin command.
+	GetLabels() *[]string
+
+	// DecideOn implements the calculation of the answer to
+	// the question.The given question contains the arguments for
```
fabianburth, line 250:
```suggestion
	// the question. The given question contains the arguments for
```

api/ocm/plugin/ppi/interface.go
```diff
@@ -211,3 +219,49 @@ type Command interface {
...
+	// the questions. There are three kinds of arguments:
+	// ArtifactQuestion, ComponentVersionQuestion and ComponentReferenceQuestion.
+	// TransferHandlerQuestions maps the question name to the used
+	// argument type.
+	DecideOn(p Plugin, question interface{}) (bool, error)
```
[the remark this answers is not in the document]
fabianburth: I think, theoretically, having dedicated `DecisionHandler` types (and dedicated `TransferHandler.RegisterDecision(DecisionHandler)` methods would be even better.

api/ocm/ref.go
```diff
@@ -145,14 +145,6 @@ func (r *RefSpec) Name() string {
...
 }
 
-func (r *RefSpec) HostPort() (string, string) {
-	i := strings.Index(r.Host, ":")
-	if i < 0 {
-		return r.Host, ""
-	}
-	return r.Host[:i], r.Host[i+1:]
-}
-
```
fabianburth, line 155: Why is this being removed? As it's an API change, we should probably mention this in the release notes.

api/ocm/tools/transfer/transfer.go
```diff
@@ -255,7 +255,7 @@ func copyVersion(printer common.Printer, log logging.Logger, hist common.History
 				hint := ocmcpi.ArtifactNameHint(a, src)
 				old, err = cur.GetResourceByIdentity(r.Meta().GetIdentity(srccd.Resources))
 
-				changed := err != nil || old.Digest == nil || !old.Digest.Equal(r.Meta().Digest)
+				changed := err != nil || old.Digest == nil || r.Meta().Digest == nil || !old.Digest.Equal(r.Meta().Digest)
```
fabianburth, line 258: why is this added? isn't this already covered by the `!old.Digest.Equal(r.Meta().Digest)` part?

api/ocm/tools/transfer/transferhandler/plugin/options.go
```diff
@@ -0,0 +1,178 @@
...
+	_ PluginNameOption            = (*Options)(nil)
+	_ TransferHandlerOption       = (*Options)(nil)
+	_ TransferHandlerConfigOption = (*Options)(nil)
+)
+
+type TransferOptionsCreator = transferhandler.SpecializedOptionsCreator[*Options, Options]
+
+func (o *Options) NewOptions() transferhandler.TransferHandlerOptions {
+	return &Options{}
+}
```
fabianburth, line 37: So, the `transferhandler.SpecializedOptionsCreator[*Options, Options]` creates a "version" of `Option` that only provides the following interface:
```go
interface {
  NewOptions() transferhandler.TransferHandlerOptions
}
```
What's the purpose of this, seeing that Options implements the interface itself?

api/ocm/tools/transfer/transferhandler/plugin/options.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/mandelsoft/goutils/errors"
+
+	"ocm.software/ocm/api/ocm/tools/transfer/transferhandler"
+	"ocm.software/ocm/api/ocm/tools/transfer/transferhandler/standard"
+	"ocm.software/ocm/api/utils/runtime"
+)
+
+func init() {
+	transferhandler.RegisterHandler(100, &TransferOptionsCreator{})
```
fabianburth, line 14:
```suggestion
	transferhandler.RegisterHandler(100, &Options{})
```

api/ocm/plugin/ppi/cmds/transferhandler/cmd.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+	}
+	opts.AddFlags(cmd.Flags())
+	return cmd
+}
+
+type Options struct {
+	Handler   string
+	Question  string
+	Arguments interface{}
```
fabianburth, line 68: `Arguments` appears to be unused, what is it supposed to do?

fabianburth:
Since I've been struggling to see the plugin architecture since it is quite scattered throughout the library, is there documentation providing an overview?
Thus, the plugin implementer side including the role of the _Plugin Descriptor_ and the plugin implementation support library (_Plugin Programming Interface (PPI)_) or a raw command implementation, as well as the plugin caller side including _Plugin Adapter_ , the registrations and the individual extension point specific plugin packages AND how all these parts fit together.

api/ocm/plugin/internal/transfer.go
```diff
@@ -0,0 +1,128 @@
...
+// For type assignments see TransferHandlerQuestions.
+type Question interface{}
+
+// ComponentVersionQuestion describes the question arguments
+// given for a component version related question.
+type ComponentVersionQuestion struct {
+	Source  SourceComponentVersion `json:"source"`
+	Target  TargetRepositorySpec   `json:"target"`
+	Options TransferOptions        `json:"options"`
+}
```
fabianburth, line 102: Significantly better than before. But I'd still like to see a correlation of the `Question` interface to the actual question types (e.g. implement methods for asserting `AsComponentVersionQuestion() (*ComponentVersionQuestion, error)`, `AsArtifactQuestion() (*ArtifactQuestion, error)`). 
WDYT?

api/ocm/plugin/ppi/utils.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"reflect"
 
+	errors2 "github.com/mandelsoft/goutils/errors"
+	"github.com/mandelsoft/goutils/generics"
 	"github.com/pkg/errors"
```
fabianburth, line 9: did you import this errors package by mistake?

api/ocm/plugin/ppi/utils.go
```diff
@@ -115,6 +119,111 @@ func (b *nameDescription) Description() string {
...
+	return d.labels
+}
+
+func NewDecisionHandlerBase(q, desc string, labels ...string) DecisionHandlerBase {
+	return DecisionHandlerBase{q, desc, generics.Pointer(slices.Clone(labels))}
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+func ComponentVersionQuestionFunv(f func(p Plugin, question *ComponentVersionQuestion) (bool, error)) func(p Plugin, question Question) (bool, error) {
```
fabianburth, line 189:
```suggestion
func ComponentVersionQuestionFunc(f func(p Plugin, question *ComponentVersionQuestion) (bool, error)) func(p Plugin, question Question) (bool, error) {
```
typo

api/ocm/plugin/ppi/utils.go
```diff
@@ -115,6 +119,111 @@ func (b *nameDescription) Description() string {
...
+}
+
+type defaultDecisionHandler struct {
+	DecisionHandlerBase
+	handler func(p Plugin, question Question) (bool, error)
+}
+
+// NewDecisionHandler provides a default decision handler based on its standard
+// fields and a handler function.
+func NewDecisionHandler(q, desc string, h func(p Plugin, question Question) (bool, error), labels ...string) DecisionHandler {
```
fabianburth, line 214: If the question types (`ComponentVersionQuestion`, `ComponentReferenceQuestion`, `ArtifactQuestion`, ...) actually implement the `Question` interface and if you define `func(p Plugin, question Question) (bool, error)` as a dedicated type, `ComponentVersionQuestionFunc `, `ComponentReferenceQuestionFunc`, `ArtifactQuestionFunc` would implement that type, which would be quite helpful when trying to implement a `NewDecisionHandler`.

api/ocm/tools/transfer/transferhandler/options.go
```diff
@@ -0,0 +1,40 @@
...
+
+//////////////////////////////////////////////////////////////////////////////
+
+// ConfigOption describes generic non-standard options.
+// Specialized option set implementation map such generic
+//
+//	config to their specialized settings. The format depends
+//
+// on the option target. For example, for spiff it is a spiff
+// script.
```
fabianburth, line 15:
```suggestion
// ConfigOption describes generic non-standard options.
// Specialized option set implementation map such generic
// config to their specialized settings. The format depends
// on the option target. For example, for spiff it is a spiff
// script.
```

the format of this comment seems broken

api/ocm/ref.go
mandelsoft: The method has been moved to the `UniformRepositorySpec`, which is an embedded field.

api/ocm/tools/transfer/transfer.go
mandelsoft: hmm, should be

api/ocm/plugin/internal/transfer.go
mandelsoft: The types are just value container, there are no methods. I could add a dummy methods to be able to find implementations.

api/ocm/tools/transfer/transferhandler/plugin/options.go
mandelsoft: you are right, the method can be removed

api/ocm/plugin/internal/transfer.go
mandelsoft: Added method QuestionArgumentsType

api/ocm/plugin/ppi/utils.go
mandelsoft: introduced type `QuestionResultFunc`

api/ocm/tools/transfer/transferhandler/plugin/options.go
mandelsoft: It is the base implementation for a specialized transfer handler, which always should be able provide an options object for its own configuration.

This is used by the heuristic, which figures out the best matching handler implementation for a given set of options.
Once determined, the final options implementation is provided by the selected handler.

mandelsoft: The only doc , so far, is the plugin reference docu generated from the ppi command framework, which describes the plugin interface.

For the ppi itself, there is only the go doc. and examples, but no closed document describing how to use ppi

hilmarf:
epic: https://github.com/open-component-model/ocm/issues/1130

api/ocm/plugin/ppi/utils.go
```diff
@@ -115,6 +118,148 @@ func (b *nameDescription) Description() string {
...
+
+func NewTransferSourceDecision(desc string, h ArtifactQuestionFunc, labels ...string) DecisionHandler {
+	return NewDecisionHandler(Q_TRANSFER_SOURCE, desc, ForArtifactQuestion(h))
+}
+
+func NewEnforceTransportDesision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
+	return NewDecisionHandler(Q_ENFORCE_TRANSPORT, desc, ForComponentReferenceQuestion(h))
+}
+
+func NewTransferversionDecision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
```
fabianburth, line 245:
```suggestion
func NewTransferVersionDecision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
```
typo

api/ocm/tools/transfer/transferhandler/options.go
```diff
@@ -0,0 +1,38 @@
...
+type ConfigOption interface {
+	SetConfig([]byte)
+	GetConfig() []byte
+}
+
+type configOption struct {
+	config []byte
+}
```
fabianburth, line 21: I feel like it is confusing that the type `configOption` is not the type that's supposed to be implementing the interface `ConfigOption`.

mandelsoft: It is supposed as option (like the other struct). In this case an option to feed the ConfigOption interface.

mandelsoft: Renamed to `ConfigOptionConsumer` and added more docu.

api/ocm/plugin/ppi/utils.go
mandelsoft: done

api/ocm/plugin/ppi/utils.go
```diff
@@ -115,6 +118,148 @@ func (b *nameDescription) Description() string {
...
+
+func NewOverwriteVersionDecision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
+	return NewDecisionHandler(Q_OVERWRITE_VERSION, desc, ForComponentReferenceQuestion(h))
+}
+
+func NewUpdateVersionDecision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
+	return NewDecisionHandler(Q_UPDATE_VERSION, desc, ForComponentReferenceQuestion(h))
+}
+
+func NewOUpdateVersionDecision(desc string, h ComponentReferenceQuestionFunc, labels ...string) DecisionHandler {
```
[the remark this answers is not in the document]
mandelsoft: no, duplicate function with typo in name. (removed)

jakobmoellerdev:
closing as we do not want to extend the existing CLI with transfer handlers anymore and are instead preparing for the move to OCM CLI v2

PR #943, opened 2024-09-30
Add a simpler digest verify function

frewilhelm:
As proposed and discussed a simpler version of `VerifyResourceDigest`.

The new name is a bit misleading and probably the name of the old function should be adjusted.

@mandelsoft What do you think? And can you please explain the behaviour of the `VerifiedStore`?

### DIFF
--- a/api/ocm/tools/signing/digest.go
+++ b/api/ocm/tools/signing/digest.go
@@ -13,6 +13,18 @@ import (
 	common "ocm.software/ocm/api/utils/misc"
 )
 
+// VerifyResourceDigestByResourceAccess verifies the digest of a resource passed by ResourceAccess.
+func VerifyResourceDigestByResourceAccess(cv ocm.ComponentVersionAccess, rAcc ocm.ResourceAccess, bacc ocm.DataAccess, ostore ...VerifiedStore) (bool, error) {
+	cd := cv.GetDescriptor()
+
+	index := cd.GetResourceIndex(rAcc.Meta())
+	if index < 0 {
+		return false, errors.ErrNotFound("resource")
+	}
+
+	return VerifyResourceDigest(cv, index, bacc, ostore...)
+}
+
 // VerifyResourceDigest verify the digest of a resource taken from a component version.
 // The data of the resources (typically after fetching the content) is given by a ocm.DataAccess.
 // The digest info is table from the resource described by a component version, which has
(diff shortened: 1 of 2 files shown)


hilmarf (approved):

api/ocm/tools/signing/digest.go
```diff
@@ -78,6 +67,28 @@ func VerifyResourceDigest(cv ocm.ComponentVersionAccess, i int, bacc ocm.DataAcc
...
+	if store != nil {
+		vcd := store.Get(cv)
+		if vcd == nil {
+			return false, fmt.Errorf("component version %s not verified", common.VersionedElementKey(cv))
+		}
+		if !vcd.Resources[i].Digest.Equal(raw.Digest) {
+			return false, fmt.Errorf("component version %s corrupted", common.VersionedElementKey(cv))
+		}
+
+		return true, nil
```
hilmarf, line 86: if the store is ok, but the cv not found, the verification should happen as if no store was provided

```suggestion
		if vcd != nil {
			return vcd.Resources[i].Digest.Equal(raw.Digest), nil
		}
```

frewilhelm:
After discussing it with @mandelsoft, we decided to add a function that takes a resource access and gets the respective index through it. Then we can call the original function again.

api/ocm/tools/signing/digest.go
```diff
@@ -51,6 +54,18 @@ func VerifyResourceDigest(cv ocm.ComponentVersionAccess, i int, bacc ocm.DataAcc
...
 
+	// Check if the resource has already been verified
+	store := general.Optional(ostore...)
+	if store != nil {
+		vcd := store.Get(cv)
+		if vcd != nil {
+			if vcd.Resources[i].Digest.Equal(raw.Digest) {
+				return true, nil
+			}
+			return false, fmt.Errorf("component version %s corrupted", common.VersionedElementKey(cv))
```
hilmarf, line 65: isn't it kind of double response the same thing? `true + no error` vs. `false + error`
What does indicate, that the verification was ok? The boolean or the error? In other words: I would not return an error, when the digests are not equal as long as no error occurred (e.g. during digest calculation or store reading or some io-error ...)

frewilhelm: yes thats funny^^

> // The function returns true if the verification has been executed. If an error occurs, or
> // the verification has been failed, an appropriate error occurs.
> // If the resource is not signature relevant (false,nil) is returned.

I will adjust it to return `true` and adjust the error ^^

hilmarf (approved):

hilmarf (approved):

fabianburth:
The function is still wrong as it does not actually verify the data access.

PR #946, opened 2024-10-01
fix + doc for VerifyResourceDigest

mandelsoft:

### DIFF
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -34,7 +34,8 @@ ifeq ("v$(JQ)","v0.0")
 endif
 
 GOLANGCILINT_VERSION := "v1.59.1"
-GOLANGCILINTV := $(shell (golangci-lint --version 2>/dev/null || echo 0.0.0) | sed 's/.*v\([0-9\.]*\) .*/\1/')
+# under MAC the version v1.59.1, on linux the v is omitted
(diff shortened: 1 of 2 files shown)


hack/Makefile
```diff
@@ -34,7 +34,7 @@ ifeq ("v$(JQ)","v0.0")
 endif
 
 GOLANGCILINT_VERSION := "v1.59.1"
-GOLANGCILINTV := $(shell (golangci-lint --version 2>/dev/null || echo 0.0.0) | sed 's/.*v\([0-9\.]*\) .*/\1/')
+GOLANGCILINTV := $(shell (golangci-lint --version 2>/dev/null || echo 0.0.0) | sed 's/.*version \([0-9\.]*\) .*/\1/')
```
frewilhelm, line 37: `golangci-lint --version` returns `golangci-lint has version v1.59.1 built...` in my case. So the `sed`-regex wouldn't trigger since its expecting something like `version 1.0.0` instead of `version v1.0.0`

mandelsoft: The version is reported via `....version 1......`. Therefore, the old regex did never match.

The comparison is done with `v$(GOLANGCILINTV)`

mandelsoft: Ok, on linux ithe version is printed without the leading v and an MAC there is a leading v,  ... great....

frewilhelm (approved):

Issue #949, opened 2024-10-02
oras transfer fails for a weird docker repository


Skarlso:
**What happened**:

```
ocm transfer component ./transport-archive ghcr.io/skarlso/ocm-transfer-test-7 -f -V
transferring version "acme.org/test:0.0.1"...
...resource 0 podinfo[ociArtifact](stefanprodan/podinfo:6.7.0)...
Error: acme.org/test:0.0.1: transferring resource 0: unable to add blob (component acme.org/test:0.0.1 resource podinfo): exploding OCI artifact resource blob ([stefanprodan/podinfo:6.7.0] namespace skarlso/ocm-transfer-test-7/stefanprodan/podinfo: transfer artifact): transferring indexed artifact sha256:0060e4fc3052c383ea920673d08388fd6aa3bfc3536932f7c08edd4e4a616520: getting layer blob sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99: failed to get blob data for artifact access with digest sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99: failed to construct data access: failed to resolve descriptor with digest sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99: HEAD "https://registry-1.docker.io/v2/stefanprodan/podinfo/manifests/sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99": response status code 500: Internal Server Error
1 versions transferred
Error: transfer finished with 1 error(s)
Error summary:
[...]

PR #950, opened 2024-10-02
fix: docker registry returning internal server error on blobs not found

Skarlso:

Fixes: https://github.com/open-component-model/ocm/issues/949

### DIFF
--- a/api/oci/extensions/repositories/ocireg/namespace.go
+++ b/api/oci/extensions/repositories/ocireg/namespace.go
@@ -54,10 +54,6 @@ func (n *NamespaceContainer) IsReadOnly() bool {
 	return n.repo.IsReadOnly()
 }
 
-func (n *NamespaceContainer) GetBlobDescriptor(digest digest.Digest) *cpi.Descriptor {
-	return nil
-}
-
 func (n *NamespaceContainer) GetBlobData(digest digest.Digest) (int64, cpi.DataAccess, error) {
 	n.repo.GetContext().Logger().Debug("getting blob", "digest", digest)
 
--- a/api/oci/extensions/repositories/ocireg/utils.go
+++ b/api/oci/extensions/repositories/ocireg/utils.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"strings"
 	"sync"
 
 	"github.com/containerd/containerd/remotes"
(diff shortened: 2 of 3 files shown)


api/oci/extensions/repositories/ocireg/namespace.go
```diff
@@ -54,10 +54,6 @@ func (n *NamespaceContainer) IsReadOnly() bool {
 	return n.repo.IsReadOnly()
 }
 
-func (n *NamespaceContainer) GetBlobDescriptor(digest digest.Digest) *cpi.Descriptor {
```
Skarlso, line 57: This was unused and didn't implement any interface methods.

api/oci/extensions/repositories/ocireg/utils.go
```diff
@@ -38,7 +39,8 @@ func NewDataAccess(repo registry.Repository, digest digest.Digest, delayed bool)
 	var reader io.ReadCloser
 	desc, err := repo.Resolve(dummyContext, digest.String())
 	if err != nil {
-		if errors.Is(err, errdef.ErrNotFound) {
+		// docker registry sometimes returns an Internal Server Error instead of NOT FOUND!
+		if errors.Is(err, errdef.ErrNotFound) || strings.Contains(err.Error(), "Internal Server Error") {
```
jakobmoellerdev, line 43: This feels like a hack for a proper retry, wdyt?

Skarlso: There is no retry. It's not an error that can go away. Docker registry replies with an internal error when you are trying to find a blob that doesn't exist.

jakobmoellerdev: Is this just a docker.io thing? If so should we maybe set this up explicitly for that registry? because all 500 feel a little strange.

Skarlso:
```
ocm transfer commontransportarchive --copy-resources ./transport-archive ghcr.io/skarlso/ocm-transfer-test-11
transferring component "acme.org/test"...
  transferring version "acme.org/test:0.0.1"...
  ...resource 0 podinfo[ociArtifact](stefanprodan/podinfo:6.2.0)...
  ...resource 1 java-sap-vcap-services[mavenPackage](com.sap.cloud.environment.servicebinding:java-sap-vcap-services:0.10.4)...
  ...resource 2 npmjs-yargs[npmPackage](yargs:17.7.2)...
  ...adding component version...
```

jakobmoellerdev (approved):
LGTM, nice fix

PR #953, opened 2024-10-08
OCI: prefer digest over tag

### LINKED ISSUE #911: OCM-CLI should prefer digest-tag for "mixed" OCI-Artefact-Tags
Widely-spread OCI-Runtimes, such as dockerd, or containerd allow usage of "mixed" OCI-Artefact-tags (tag + digest -  `<repository>:<tag>@<digest>`). Such tags are interpreted as digest-tags (i.e. symbolic tag is ignored). Motivation for using such "mixed" tags is to keep human-friendly tags, while still benefit from content-addressing.

When replicating OCM-Components w/ OCI-Artefacts w/ such tags (`ocm transfer`), OCM-CLI seems to prefer symbolic tag, and ignore digest, which consequently breaks if "mixed-tags" are converted to just digest-tags in an "out-of-bands"-replication-process.

Hence, I suggest to change OCM-CLI's behaviour like so:

If resource-access's imageReference attribute contains "mixed tag":
- digest-tag should be honoured for _retrieving_ the artefact (ocm-cli should drop symbolic part for retrieval)
- mixed-tag should be retained for re-upload

It might be a good idea to make the latter behaviour configurable, as there might be target-oci-registries that will not accept mixed-tags. Genereally-speaking, conversion of tags between symbolic <-> mixed <-> digest might be offered as an option for `ocm transfer` cmd. Might also be covered via  #902 .

mandelsoft:

If in a OCI ref both, the tag and the digest is given, the tag is preferred, so far. BUt the digest should always have precedence.

Fixes #911

### DIFF
--- a/api/ocm/extensions/download/handlers/ocirepo/handler.go
+++ b/api/ocm/extensions/download/handlers/ocirepo/handler.go
@@ -55,7 +55,7 @@ func (h *handler) Download(p common.Printer, racc cpi.ResourceAccess, path strin
 
 	var repo oci.Repository
 
-	var version string = "latest"
+	var tag string
 
 	aspec := m.AccessSpec()
 	namespace := racc.ReferenceHint()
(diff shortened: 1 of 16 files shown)


api/ocm/extensions/download/handlers/ocirepo/handler.go
```diff
@@ -113,23 +124,23 @@ func (h *handler) Download(p common.Printer, racc cpi.ResourceAccess, path strin
 	}
 	log.Debug("using artifact spec", "spec", artspec.String())
 	if artspec.Digest != nil {
-		return true, "", fmt.Errorf("digest no possible for target")
+		return true, "", fmt.Errorf("digest not possible for target")
```
jakobmoellerdev, line 127: could you quickly explain to me why a digest isnt possible? because you cant upload by digest?

jakobmoellerdev (commented):
LGTM, just want to have some explanations on the changes so I can understand better

api/ocm/extensions/download/handlers/ocirepo/handler.go
mandelsoft: the digest is defined by the provided content, it is not given by the upload target spec.

jakobmoellerdev (approved):
thanks for the addition of test cases which cover the relevant tag/digest combinations

PR #954, opened 2024-10-10
CLI:  fix recursive mode for ocm hash component

mandelsoft:

In recursive mode of `ocm hash components`, the closure calculation is done in the processing chain after
the mapping to the hash object. This cause type assertion panics.

Additionally, the recalculating hashes for a component version using a shared state, also a panic occurs, because it has
already been calculated as part of a closure.

Both problems are fixed with this PR.

Then, it adds some more flexibility to the input handling, enabling the use of the constructor processing in 
a more flexible way for other programs besides the usual cli commands.

### DIFF
--- a/cmds/ocm/commands/ocmcmds/common/addhdlrs/interface.go
+++ b/cmds/ocm/commands/ocmcmds/common/addhdlrs/interface.go
@@ -17,9 +17,14 @@ import (
 // a content based element (sources or resources).
 // It is either an input or access specification.
 type ResourceInput struct {
-	Access *cpi.GenericAccessSpec `json:"access"`
+	SourceFile string                 `json:"sourceFile,omitempty"`
+	Access     *cpi.GenericAccessSpec `json:"access"`
 	// Input  *inputs.BlobInput                `json:"input,omitempty"`
 	Input *inputs.GenericInputSpec `json:"input,omitempty"`
+}
+
+func (r *ResourceInput) SetSourceFile(s string) {
+	r.SourceFile = s
 }
 
 // ElementSpecHandler is the interface for a handler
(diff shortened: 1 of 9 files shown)


cmds/ocm/commands/ocmcmds/common/addhdlrs/interface.go
```diff
@@ -17,11 +17,16 @@ import (
 // a content based element (sources or resources).
 // It is either an input or access specification.
 type ResourceInput struct {
-	Access *cpi.GenericAccessSpec `json:"access"`
+	SourceFile string                 `json:"sourceFile,omitempty"`
```
fabianburth, line 20: add comment explaining why this is needed

cmds/ocm/commands/ocmcmds/components/hash/cmd.go
```diff
@@ -102,8 +102,8 @@ func TableOutput(opts *output.Options, h *action, mapping processing.MappingFunc
 	def := &output.TableOutput{
 		Headers: output.Fields("COMPONENT", "VERSION", "HASH", wide),
 		Options: opts,
-		Chain:   comphdlr.Sort.Map(h.digester),
-		Mapping: mapping,
+		Chain:   comphdlr.Sort,
+		Mapping: processing.MappingSequence(h.digester, mapping),
```
fabianburth, line 106: The current logic needs rework urgently, but the since the fix is simple, it's alright for now.

cmds/ocm/commands/ocmcmds/common/cmds/signing/cmd.go
```diff
@@ -139,6 +139,8 @@ func NewAction(desc []string, ctx ocm.Context, p common.Printer, sopts *signing.
 func (a *action) Digest(o *comphdlr.Object) (*metav1.DigestSpec, *compdesc.ComponentDescriptor, error) {
 	sopts := *a.sopts
 	sopts.Resolver = resolvers.NewCompoundResolver(o.Repository, a.sopts.Resolver)
+
+	a.state.Get(common.VersionedElementKey(o.ComponentVersion))
```
fabianburth, line 143: I guess this can be removed.

fabianburth (approved):

PR #955, opened 2024-10-11
feat: add function to satisfy accessMethodView interface

frewilhelm:
Otherwise the type-assertion for redirected access methods won't be `ok` for non-local oci-artifacts, see for example:

https://github.com/open-component-model/ocm/blob/main/api/ocm/tools/signing/digest.go#L77
v
https://github.com/open-component-model/ocm/blob/main/api/ocm/extensions/digester/digesters/artifact/digester.go#L141
v 
https://github.com/open-component-model/ocm/blob/main/api/ocm/cpi/accspeccpi/methodview.go#L142
v
not `ok` even though it should work.

### DIFF
--- a/api/ocm/tools/signing/digest.go
+++ b/api/ocm/tools/signing/digest.go
@@ -107,6 +107,8 @@ type redirectedAccessMethod struct {
 	acc ocm.DataAccess
 }
 
+var _ accspeccpi.AccessMethodView = (*redirectedAccessMethod)(nil)
+
 func NewRedirectedAccessMethod(m ocm.AccessMethod, bacc ocm.DataAccess) ocm.AccessMethod {
 	return &redirectedAccessMethod{m, bacc}
 }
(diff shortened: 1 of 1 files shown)


api/ocm/tools/signing/digest.go
```diff
@@ -110,6 +111,15 @@ func NewRedirectedAccessMethod(m ocm.AccessMethod, bacc ocm.DataAccess) ocm.Acce
 	return &redirectedAccessMethod{m, bacc}
 }
```
fabianburth, line 113:
```suggestion
var _ accspeccpi.AccessMethodView = (*redirectedAccessMethod)(nil)
```

Ideally, you add this right below the `redirectedAccessMethod` type definition.

fabianburth (changes requested):

api/ocm/tools/signing/digest.go
frewilhelm: done

fabianburth (approved):

PR #958, opened 2024-10-14
Support standard object types for command plugins + options for templater types

mandelsoft:
- Support standard object types for command plugins 
  The CLI supports verbs on objects (e.g, get component). For such object names aliases are used 
  (e.g. for componentversions, it is possible to use component or cv, instead)
  A CLI command provided by a plugin may use such a standard object name. The new integration takes the aliases provided
  by the CLI into account.
- Options for templater types
  Templaters are used by the CLI to process the constructor files, but possible also by direct lib users.
  So far, templaters are created for a value set, but without other arbitrary type specific optional parameters.
  The templater configuration now offers settings passed to the templater constructor. This is used by spiff
  to enable or restrict the OS mode. By default, file system access is enabled,. OS access is enabled if no virtual filesystem
  is used. With this configuration option, the user of the templater may configure tis behaviour.

### DIFF
--- a/cmds/ocm/commands/ocmcmds/names/names.go
+++ b/cmds/ocm/commands/ocmcmds/names/names.go
@@ -18,3 +18,32 @@ var (
 	PubSub                 = []string{"pubsub", "ps"}
 	Verified               = []string{"verified"}
 )
+
+var Aliases = map[string][]string{}
+
+func init() {
+	add(
+		ComponentArchive,
+		CommonTransportArchive,
+		Components,
+		CLI,
+		Configuration,
+		ResourceConfig,
+		SourceConfig,
+		Resources,
+		Sources,
+		References,
+		Versions,
+		Plugins,
+		Action,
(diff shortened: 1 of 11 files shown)


fabianburth (changes requested):
The PR description does not explain, why we need it. Please provide more information on that?

In regards to the templater, for which commands is it used on the command line?

cmds/ocm/commands/ocmcmds/names/names.go
```diff
@@ -18,3 +18,15 @@ var (
 	PubSub                 = []string{"pubsub", "ps"}
 	Verified               = []string{"verified"}
 )
+
+var Aliases = map[string][]string{}
+
+func init() {
+	add(Components, Resources, Sources)
```
fabianburth, line 25: Why are only `Components`, `Resources` and `Sources` added?

fabianburth (approved):

PR #964, opened 2024-10-15
Helm downloader registration handler

fabianburth:
Registering the Registration Handler of the Helm Downloader Handler allows to register the Helm Download Handler for other artifact types (aka. resource types) than `helmChart`, for which it is configured by default.

This configuration can be done through adding this section to the ocm config file (or passing such a config directly). 

```yaml
        type: downloader.ocm.config.ocm.software
        description: "my standard download handler configuration"
        handlers:
          - name: helm/artifact
            artifactType: NotHelmChart
            mimeType: application/vnd.oci.image.manifest.v1+json
```


api/ocm/extensions/download/handlers/helm/registration.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"ocm.software/ocm/api/oci/artdesc"
+	"ocm.software/ocm/api/oci/extensions/repositories/artifactset"
+	"ocm.software/ocm/api/ocm/cpi"
+	downloadhandlers "ocm.software/ocm/api/ocm/extensions/download"
+	"ocm.software/ocm/api/utils/listformat"
+	"ocm.software/ocm/api/utils/registrations"
+)
+
+const PATH = "helm/artifact"
```
hilmarf, line 18: I hope we don't lose control for so many different `const` values...
https://github.com/open-component-model/ocm/blob/main/cmds/ocm/commands/ocmcmds/common/inputs/types/helm/type.go#L7
https://github.com/open-component-model/ocm/blob/main/api/ocm/elements/artifactaccess/helmaccess/resource.go#L12
https://github.com/open-component-model/ocm/blob/main/api/ocm/extensions/artifacttypes/const.go#L16

hilmarf (dismissed):

hilmarf (approved):

PR #974, opened 2024-10-17
fix: GHA - .github/workflows/flake_vendorhash.yaml

hilmarf:

The current action ends up in an endless loop, since it updates the PR which triggered it.

With v7 of `peter-evans/create-pull-request` we're now able to easily create PRs with multiple files and a signed commit!


jakobmoellerdev (approved):
very cool correction!

PR #977, opened 2024-10-18
chore: add release notes for v0.16.1

Skarlso:


jakobmoellerdev (changes requested):
Thats incorrect :D 

1. we need to create a patch with https://github.com/open-component-model/ocm/actions/workflows/release-branch.yaml
2. We need to point it to a tag v0.16.1 and manually adjust the version file
3. We then need to press release on that?

hilmarf:
> Thats incorrect :D
> 
> 1. we need to create a patch with https://github.com/open-component-model/ocm/actions/workflows/release-branch.yaml
> 2. We need to point it to a tag v0.16.1 and manually adjust the version file
> 3. We then need to press release on that?

First someone would need to create a branch based upon tag v0.16.0, then do cherry-pick the fix into the branch and then create a release based upon that new branch. That's how I understood the release process. Am I right @fabianburth / @mandelsoft ?

PR #987, opened 2024-10-18
!OCI Annotations change in Component Descriptor for Component Versions: `ocm-artifact` -> `software.ocm.artifact` and introduce `software.ocm.creator`

mandelsoft:

For OCM repositories based on OCI registries, the manifest now gets an annotation `software.ocm.componentversion` with the component version it represents.

The image spec (https://github.com/opencontainers/image-spec/blob/main/annotations.md) state, that annotation names should use a reverse DNS domain prefix.

Therefore we should adapt the names of the already existing layer annotations accordingly. This PR prepends the actuals names with `software.ocm.".

The old layer annotation changes from `ocm-artifact` -> `software.ocm.artifact`.

Fixes https://github.com/open-component-model/ocm-project/issues/190


api/ocm/extensions/repositories/genericocireg/annotation_test.go
```diff
@@ -63,6 +64,9 @@ var _ = Describe("Transfer handler", func() {
 		Expect(v.GetDescriptor()).To(YAMLEqual(`
   mediaType: application/vnd.oci.image.manifest.v1+json
   schemaVersion: 2
+  annotations:
```
jakobmoellerdev, line 67: this should not be part of this PR as I mentioned in our sync as now the title and scope of the issue no longer fit IMO

jakobmoellerdev: I suggest you draw these commits out again and create a separate PR, its not necessary to build on this one

docs/reference/ocm_add_resource-configuration.md
```diff
@@ -607,60 +607,6 @@ The access method specification can be put below the <code>access</code> field.
 If always requires the field <code>type</code> describing the kind and version
 shown below.
 
-- Access type <code>S3</code>
```
jakobmoellerdev, line 610: This is also not part of this issue, please create a separate PR.

jakobmoellerdev (approved):
LGTM for me, feel free to move out of draft.

Issue #990, opened 2024-10-21
Also make generate after version bump in Release Job to correctly update README.md with component version references


jakobmoellerdev:
**What happened**:

https://github.com/open-component-model/ocm/actions/runs/11407269902/job/31742898240?pr=987

causes issues after release due to check diff failing

**What you expected to happen**:

https://github.com/open-component-model/ocm/blob/main/.github/workflows/release.yaml#L286 should also contain a make generate call to refresh the readme.

**How to reproduce it (as minimally and precisely as possible)**:

1. Release
2. Observe failing PR

**Anything else we need to know**:

**Environment**:

PR #992, opened 2024-10-21
chore(release): also call make generate to update files

jakobmoellerdev:
This allows us to also generate the README.md files automatically so that they contain correct references

#990

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -282,8 +282,9 @@ jobs:
         git checkout ${GITHUB_REF#refs/heads/}
         v="$(go run ./api/version/generate bump-version)"
         echo "$v" > VERSION
-        git add VERSION
-        git commit -m "Update version file to $v"
+        make generate
+        git add --all
+        git commit -m "Update version to $v"
         git push origin ${GITHUB_REF#refs/heads/}
         echo "Next branch version is $v"
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -282,6 +282,7 @@ jobs:
         git checkout ${GITHUB_REF#refs/heads/}
         v="$(go run ./api/version/generate bump-version)"
         echo "$v" > VERSION
+        # Trigger a bump of any potential files that depend on a new version
         make generate
         git add --all
         git commit -m "Update version to $v"


.github/workflows/release.yaml
```diff
@@ -278,10 +278,13 @@ jobs:
...
       if: inputs.release_candidate == false
       run: |
+        # Trigger a bump of the version file
         set -e
         git checkout ${GITHUB_REF#refs/heads/}
         v="$(go run ./api/version/generate bump-version)"
         echo "$v" > VERSION
+        # Trigger a bump of any potential files that depend on a new version
+        make generate
         git add VERSION
```
hilmarf, line 288: I guess you'll need to add the generated markdown files here as well

jakobmoellerdev: Thanks for the correction, changed the add to --all since I dont know in advance which files changed

hilmarf (dismissed):

.github/workflows/buildcomponents.yaml
```diff
@@ -36,14 +36,12 @@ jobs:
...
         with:
           path: |
             ${{ env.go_cache }}
             ${{ env.go_modcache }}
-          key: ${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ hashFiles('**/go.mod') }}
-          restore-keys: |
-            ${{ env.cache_name }}-${{ runner.os }}-go-
+          key: ${{ env.cache_name }}-${{ runner.os }}-go-
         env:
           cache_name: buildcomponents-go-cache
```
hilmarf, line 46:
```suggestion
          cache_name: run-tests-go-cache # needs to be the same key in the end as in the build step
```

or am I wrong?

jakobmoellerdev: This is still on the old setup so its a bit of a mess, but since https://github.com/open-component-model/ocm/pull/996 is decoupled from this one, it would still have its own cache. (thats why its buildcomponents, not run-tests)

jakobmoellerdev: Just read through this again and youre right even within this PR it doesnt make sense. This PR shouldnt have this code for caching in the first place.

Skarlso:
I don't like this idea. All PRs should generate content. This potentially will leave a dirty branch or a dirty main that people don't expect. It could potentially could leave a branch with data that the developers are unaware of and had not committed or created.

I know that it's _only_ generated data, nevertheless it's unexpected code that we potentially are unaware off. I don't like just plain committing generated code into the release especially if we had guards against merging prs that have ungenerated modifications in them.

Skarlso:
For now, this is fine though. I realise that there are no other options in this case since the current automation will use the version as a stage point.

We'll figure something out. :D

Skarlso (approved):

jakobmoellerdev:
Ill create a follow up because you still have a point. We need to make sure this generation doesnt fall on our face.

PR #996, opened 2024-10-21
chore(ci): various optimizations for build processing, caching and concurrency

### LINKED ISSUE #999: Extreme Cache Thrashing in our Github actions
**What happened**:

When we separated the golang build caches per step, we exceeded our 10Gi storage limit by far more than what I thought. Our build cache is ~4Gi large just for the Multi-Arch build, and that leads to a daily cache usage of about 50Gi (when the limit is 10Gi). This means every night the caches are reset and we have long running builds again.

We should dumb down the caches to only use one or 2 storage locations (one for the multiarch build and one for the golang tests + amd64 linux build combination - which are compiled separately and would otherwise be hard to cache) so that we do not exceed our cache limits. Similarly we should not cache in PRs but only consume the module and build cache from main.

Just watch on the cache usage and the runtimes every morning. You can also check the org logs if you have access.

**Environment**:

jakobmoellerdev:
Our caches are getting too big:

```
Approaching total cache storage limit (58.09 GB of 10 GB Used)
```

On top we dont make use of makes inbuilt parallelism (`-j`) and also we do not make use of shells forking mechanism for concurrent execution (`&` execution chain followed by `wait`)

Limits the cache save to only the main branch in the build step, all other branches/PRs and Jobs will reuse that cache instead of computing their own

What does this mean:

PRO:
- only one cache that should roundabout have 4 (build files for all platforms) + 2 (build+test files for unit tests on amd64) Gi of modules and builds that get updated on every commit to main
- The build cache is likely unaffected or largely still valid for most if not every PR. This leads to compile time improvements upwards of 600% depending on PR size without killing our cache usage.

CON:
- a rebuild or dependency changes will be able to "invalidate" that cache and could lead to longer build times faster. However in general, most builds should still perform the same. If the go.mod / go.sum is not touched, the build cache will be restored for all files except the ones touched or impacted by the PR. 

fix #999 #1000


.github/workflows/push_ocm.yaml
```diff
@@ -148,3 +153,19 @@ jobs:
         skipIfReleaseExists: false
         body: |
           holds always the latest ocm-cli binaries
+
+    # This step will only reuse the go mod and build cache from main made during the Build,
+    # see push_ocm.yaml => "ocm-cli-latest" Job
+    # This means it never caches by itself and PRs cannot cause cache pollution / thrashing
+    # This is because we have huge storage requirements for our cache because of the mass of dependencies
```
hilmarf, line 160: copy & paste documentation? Seems not to fit to the actual doing.

.github/workflows/push_ocm.yaml
```diff
@@ -148,3 +153,19 @@ jobs:
...
+    - name: Save Cache of Build (only on main)
+      id: cache-golang-save
+      uses: actions/cache/save@v4 # Only Restore, not build another cache (too big)
```
hilmarf, line 163:
```suggestion
      uses: actions/cache/save@v4 # Only save build cache once
```

.github/workflows/buildcomponents.yaml
```diff
@@ -36,14 +36,12 @@ jobs:
...
         with:
           path: |
             ${{ env.go_cache }}
             ${{ env.go_modcache }}
-          key: ${{ env.cache_name }}-${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-${{ hashFiles('**/go.mod') }}
-          restore-keys: |
-            ${{ env.cache_name }}-${{ runner.os }}-go-
+          key: ${{ env.cache_name }}-${{ runner.os }}-go-
         env:
           cache_name: buildcomponents-go-cache
```
hilmarf, line 46: looks odd... at least different to the other 'restore' cases...

jakobmoellerdev: Youre right, wrong store

jakobmoellerdev:
@hilmarf could you take another look here

.github/workflows/lint_and_test.yaml
```diff
@@ -41,10 +49,32 @@ jobs:
...
             ${{ env.cache_name }}-${{ runner.os }}-go-
         env:
-          cache_name: run-tests-go-cache
+          cache_name: run-tests-go-cache # needs to be the same key in the end as in the build step
 
+      - name: Build
+        run: make build -j
       - name: Test
-        run: make build install-requirements test
+        run: make install-requirements test
```
hilmarf, line 57: just from reading perspective this looks a bit odd... why are the requirements installed AFTER the build? Are those requirements only required for testing?

.github/workflows/lint_and_test.yaml
```diff
@@ -41,10 +49,32 @@ jobs:
...
+      # see lint_and_test.yaml => "test" Job
+      # This means it never caches by itself and PRs cannot cause cache pollution / thrashing
+      # This is because we have huge storage requirements for our cache because of the mass of dependencies
+      #
+      # NOTE: This is different from our regular build cache (which contains all archs and is built in a different job)
+      # This is because it requires caching of test dependencies, which are compiled only for linux-amd64 for test runs in CI.
+      - name: Save Cache of Build (only on main)
+        id: cache-golang-save
+        if: github.ref == 'refs/heads/main' # Only run on main, never in PR
+        uses: actions/cache/save@v4 # Only Restore, not build another cache (too big)
```
hilmarf, line 69:
```suggestion
        uses: actions/cache/save@v4 # Only saves cache build-test (linux-amd64)
```

.github/workflows/lint_and_test.yaml
```diff
@@ -41,10 +49,32 @@ jobs:
...
+        run: make build -j
       - name: Test
-        run: make build install-requirements test
+        run: make install-requirements test
+
+      # This step will only reuse the go mod and build cache from main made during the Build,
+      # see lint_and_test.yaml => "test" Job
+      # This means it never caches by itself and PRs cannot cause cache pollution / thrashing
+      # This is because we have huge storage requirements for our cache because of the mass of dependencies
+      #
```
hilmarf, line 63:
```suggestion
```

hilmarf:
General question: what do we want to build as fast as possible? (What requires really the best cache setup?)

- Release builds?
- PR builds?
  - Dependency updates?
  - Feature builds?
  - Bug fixes?
- Hotfixes?
- ... ?

.github/workflows/lint_and_test.yaml
jakobmoellerdev: that is a good question. Basically from what I understand these requirements are only for testing (e.g. the vault CLI), but not for building the binary

hilmarf: should we then rename that make target to: 'install-test-requirements'?

jakobmoellerdev:
> General question: what do we want to build as fast as possible? (What requires really the best cache setup?)
> 
> * Release builds?
> * PR builds?
>   
>   * Dependency updates?
>   * Feature builds?
>   * Bug fixes?
> * Hotfixes?
> * ... ?


All build steps using go dependencies should be accelerated as much as possible. I would like to keep build times in the low 5 minutes, so also all PR builds.

However due to our exorbitantly large cache, the tradeoff is to reuse only one build cache from release builds as well as test steps on main.

.github/workflows/lint_and_test.yaml
jakobmoellerdev: the requirements are unfortunately shared between other jobs as well so it will be hard to justify that. (e.g. make generate)

fabianburth: I agree that its not that simple, since the "requirements" also include e.g. golang-ci-lint. Theoretically, I guess the best way would be to split the install-requirements target. And the test target would depend on a install-requirements. But that's probably out of scope for this PR.

hilmarf (approved):

PR #1001, opened 2024-10-22
Fix/publish release

hilmarf:

Latest releases weren't published for all cases:

- https://github.com/open-component-model/ocm/actions/runs/11404973209/job/31735340051 relies on release notes being present in git-repo
- https://github.com/open-component-model/ocm/actions/runs/11404973209/job/31735340433 version is not set properly


jakobmoellerdev (approved):
LGTM, but that should probably be tracked as part of the Release Issue

PR #1002, opened 2024-10-22
chore(ci): add stale bot for issues

### LINKED ISSUE #285: Requesting a working guide for modelling multi component and its deployment via ocm-controller.
- Needed the example guide or documentation, The documentation should describe, how to model a component consisting of several child components.
- It should be added in https://ocm.software/docs/guides/getting-started-with-ocm-flux/ .

**Why is this needed**:
-  For the deployment of multiple OCM components with root - reference (parent-child) approach  .
-  We need example for better understanding the relations between parent and child component through "reference.yaml" .
-  Using root - reference approach , there are separate component descriptors has been created for each component , and there is a component descriptors for parent as well, we need to understand how it is getting referenced .
-  And how ocm - controller will pick the child components , if parent componentversion.yaml is getting deployed on cluster ? .

jakobmoellerdev:

Adds a stale bot configuration

fix [#285](https://github.com/open-component-model/ocm-project/issues/285)

fyi @morri-son

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,23 @@
+name: 'Mark stale issues and PRs'
+on:
+  schedule:
+    - cron: '7 1 * * *'
+  workflow_dispatch:
+    inputs:
+
+permissions:
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue was marked as stale because it has not had recent activity.'
+          stale-issue-label: 'lifecycle/stale'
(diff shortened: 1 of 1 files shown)


morri-son:
@jakobmoellerdev just fixes half of [#285 ](https://github.com/open-component-model/ocm-project/issues/285). Is this enough for us and we don't need a rotten state?

jakobmoellerdev:
Discussed Async:
1. We want 90 days stale interval
2. we start with only stale and add rotten later if we need it

morri-son (approved):
lgtm

hilmarf:
@jakobmoellerdev / @morri-son - should we move this action to our central place:
https://github.com/open-component-model/.github/tree/main/.github/workflows and then call(reuse) it from every repo, where we want to clean/mark issues automatically?

see sample usage: https://github.com/open-component-model/ocm-website/blob/main/.github/workflows/verify.yml#L11

jakobmoellerdev:
yes, great idea. I'll prepare a PR

PR #1004, opened 2024-10-22
chore(ci): make sure we use candidates and branches by default

### LINKED ISSUE #997: Build Release Candidates and Nightlies

- Build nightly candidate builds that can be used for releases and testing e2e later on
- Build release candidates before publishing a release that can be distributed before the actual release and tested in early adopter environments

**Why is this needed**:

Our current release infrastructure is too fragile and directly releasing has proven fatal for our last release in v0.16 , 0.15 and 0.14 releases are better but still fragile.

jakobmoellerdev:

partially resolves #997 by being able to default to candidates, releaes branches and adding a small documentation paragraph

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,18 +6,18 @@ on:
       release_candidate:
         type: boolean
         description: "Release Candidate"
-        required: false
-        default: false
+        required: true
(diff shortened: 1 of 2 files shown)


morri-son (approved):
lgtm.

PR #1006, opened 2024-10-22
fix!(release): retract 0.16.0-z

jakobmoellerdev:

see #293 for context, causing us to redact 0.16

Redacts version 0.16 in its entirety on the next release as we do not want consume ORAS as is


morri-son (approved):
lgtm

PR #1008, opened 2024-10-23
chore(ci): use reusable stale bot configuration

jakobmoellerdev:

Changes the stale configuration to https://github.com/open-component-model/.github/blob/main/.github/workflows/stale.yml


hilmarf (approved):

hilmarf:
unfortunately we still need to add this action now to all the other repos as well... especially ocm-project

PR #1010, opened 2024-10-23
chore(ci): remove i386, optimize latest build and add various comments

jakobmoellerdev:

Explicitly marks win builds to only build for amd64 / arm64 so that we exclude i386 which is rarely used.

fix https://github.com/open-component-model/ocm-project/issues/261

### DIFF
--- a/.github/workflows/publish-to-other-than-github.yaml
+++ b/.github/workflows/publish-to-other-than-github.yaml
@@ -1,4 +1,11 @@
-name: Publish Release
+# This publish step takes care of some (but not all ;) ) of the package registries
+# that we think might be useful for people to consume.
+#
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-to-other-than-github.yaml
```diff
@@ -1,4 +1,11 @@
-name: Publish Release
+# This publish step takes care of some (but not all ;) ) of the package registries
+# that we think might be useful for people to consume.
+#
+# Other package registries might also be pushed in the goreleaser step (see publish-latest.yaml) and are configured
```
hilmarf, line 4:
```suggestion
# Other package registries might also be pushed in the goreleaser step (see release.yaml) and are configured
```

hilmarf (approved):

PR #1014, opened 2024-10-25
chore: makefile adjustment

fabianburth:
Small fix and adjustments in Makefile to avoid future potential confusion.


hack/Makefile
```diff
@@ -97,18 +97,18 @@ go-bindata:
...
-	fi
+		chmod a+x $(LOCALBIN)/vault;
+endif
 
 .PHONY: oci-registry
 oci-registry:
-	@if [ "$(OCI_REGISTRY)" != "$(OCI_REGISTRY_VERSION)" ]; then \
-		go install -v github.com/distribution/distribution/v3/cmd/registry@v3.0.0-alpha.1; \
-	fi
+ifeq (,$(findstring $(OCI_REGISTRY_VERSION), $(OCI_REGISTRY)))
```
hilmarf, line 109: is the `,` correct? or should be removed?

```suggestion
ifeq ($(findstring $(OCI_REGISTRY_VERSION), $(OCI_REGISTRY)))
```

hilmarf (approved):

Makefile
```diff
@@ -74,10 +74,11 @@ check-and-fix:
 force-test:
 	@go test --count=1 $(EFFECTIVE_DIRECTORIES)
 
+TESTFLAGS = --tags=integration
 .PHONY: test
 test:
 	@echo "> Run Tests"
-	@go test  --tags=integration $(EFFECTIVE_DIRECTORIES)
+	go test  $(TESTFLAGS) $(EFFECTIVE_DIRECTORIES)
```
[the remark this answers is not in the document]
frewilhelm: `@` omits the implicit `echo` of the command that is executed (`make` "feature").

`@go test ...` displays

```
<go test output>
```

`go test ...` displays
```
go test ...
<go test output>
```

hack/Makefile
fabianburth: This is correct, although i agree, the syntax looks weird.
If `$OCI_REGISTRY` contains `$(OCI_REGISTRY_VERSION)`, [findstring](https://www.gnu.org/software/make/manual/make.html#index-findstring) returns the value of `$OCI_REGISTRY`. Otherwise, findstring returns an empty string.
What this statement this is to check that findstring returned an empty string (thus, we have to update the version).

Skarlso (approved):

PR #1020, opened 2024-10-28
Upcoming PRs built on transfer handler plugin pr (signing plugin + input plugin)

mandelsoft:


jakobmoellerdev:
closing as we do not want to extend the existing CLI with transfer handlers anymore and are instead preparing for the move to OCM CLI v2

PR #1026, opened 2024-10-29
fix implicit version handling of an artifact identity

mandelsoft:
The defaulting of the extraIdentity did only work, if there was an identity map already set.

So far the version was implicutly added to the extraIdentity of a CV if the rest is not 
unique. In the future the extraIdentity should be explicitly set to be unique. Therefore,
the version was now implicitly added to the extraIdentity. This has two problems:
- it was only done, if there was already an identity map set,
- it was always done, when serializing a CD

To be comparable with older signed component versions,  such defaulting may only be done
if the content of a component version is changed, otherwise the signature would be brocken.

### DIFF
--- a/api/helper/builder/ocm_identity.go
+++ b/api/helper/builder/ocm_identity.go
@@ -1,5 +1,9 @@
 package builder
 
+import (
+	metav1 "ocm.software/ocm/api/ocm/compdesc/meta/v1"
+)
+
 const T_OCMMETA = "element with metadata"
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -8,6 +12,18 @@ func (b *Builder) ExtraIdentity(name string, value string) {
 	b.expect(b.ocm_meta, T_OCMMETA)
 
 	b.ocm_meta.ExtraIdentity.Set(name, value)
+}
+
+func (b *Builder) ExtraIdentities(extras ...string) {
+	b.expect(b.ocm_meta, T_OCMMETA)
+
+	id := metav1.NewExtraIdentity(extras...)
+	if b.ocm_meta.ExtraIdentity == nil {
+		b.ocm_meta.ExtraIdentity = metav1.Identity{}
+	}
+	for k, v := range id {
+		b.ocm_meta.ExtraIdentity.Set(k, v)
+	}
 }
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
+++ b/cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
@@ -23,19 +23,19 @@ var (
 )
 
 func (o *Options) AddFlags(fs *pflag.FlagSet) {
-	f := fs.Lookup("replace")
+	o.addBoolFlag(fs, &o.Replace, "replace", "R", false, "replace existing elements")
+	o.addBoolFlag(fs, &o.PreserveSignature, "preserve-signature", "P", false, "preserve existing signatures")
+}
+
+func (o *Options) addBoolFlag(fs *pflag.FlagSet, p *bool, long string, short string, def bool, usage string) {
+	f := fs.Lookup(long)
 	if f != nil {
-		if f.Value.Type() == "bool" {
-			return
-		}
-		if f == nil {
-			fs.BoolVarP(&o.Replace, "replace", "R", false, "replace existing elements")
+		if f.Value.Type() != "bool" {
+			f = nil
 		}
 	}
-
-	f = fs.Lookup("preserve-signature")
 	if f == nil {
-		fs.BoolVarP(&o.PreserveSignature, "preserve-signature", "P", false, "preserve existing signatures")
+		fs.BoolVarP(p, long, short, def, usage)
 	}
 }
 
--- a/cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
+++ b/cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
@@ -1,16 +1,23 @@
 package addhdlrs
 
 import (
+	"github.com/mandelsoft/goutils/generics"
 	"github.com/spf13/pflag"
 
 	"ocm.software/ocm/api/ocm"
 )
 
 type Options struct {
-	Replace bool
+	Replace           bool
+	PreserveSignature bool
 }
 
-var _ ocm.ModificationOption = (*Options)(nil)
+var (
+	_ ocm.ModificationOption        = (*Options)(nil)
+	_ ocm.ElementModificationOption = (*Options)(nil)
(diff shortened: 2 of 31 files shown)


jakobmoellerdev:
> A second fix is included for the creation of an artifact set for an OCI artifact access.
The switch from tag to digest priority resulted in an artifact set not containing all the manifest
meta data. This is fixed by keeping both parts (tag + digest) and extracting the optional
tag for creating the manifest meta data.

Would you mind separating this into 
1. a separate commit
2. a new PR

I will continue reviewing the existing fix for the implicit version handling

api/helper/builder/ocm_identity.go
```diff
@@ -10,6 +14,18 @@ func (b *Builder) ExtraIdentity(name string, value string) {
 	b.ocm_meta.ExtraIdentity.Set(name, value)
 }
 
+func (b *Builder) ExtraIdentities(extras ...string) {
+	b.expect(b.ocm_meta, T_OCMMETA)
+
+	id := metav1.NewExtraIdentity(extras...)
+	if b.ocm_meta.ExtraIdentity == nil {
```
jakobmoellerdev, line 21: I just noticed this, but why do all fields not correlate to the idiomatic naming? such as 

```suggestion
	if b.ocmMeta.ExtraIdentity...
```

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
```diff
@@ -8,30 +8,51 @@ import (
 )
 
 type Options struct {
-	Replace bool
+	Replace           bool
+	PreserveSignature bool
```
jakobmoellerdev, line 12: These Options need some docs I think for incode referencing

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
```diff
@@ -8,30 +8,51 @@ import (
...
+)
 
 func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	f := fs.Lookup("replace")
-	if f != nil {
-		if bp := generics.Cast[*bool](f.Value); bp != nil {
-			return
-		}
+	if f == nil {
+		fs.BoolVarP(&o.Replace, "replace", "R", false, "replace existing elements")
```
jakobmoellerdev, line 25: Why do you use capslock for R and P? Is there already a shorthand for lower case given?

jakobmoellerdev: I think the doc "replace existing elements" needs a touchup to describe when to use this flag, not just what it does

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
```diff
@@ -8,30 +8,51 @@ import (
...
+	}
+
+	f = fs.Lookup("preserve-signature")
+	if f == nil {
+		fs.BoolVarP(&o.PreserveSignature, "preserve-signature", "P", false, "preserve existing signatures")
```
jakobmoellerdev, line 30:
```suggestion
		fs.BoolVarP(&o.PreserveSignature, "preserve-signature", "P", false, "preserve existing signatures")
```
similar to above. I think we should write here when to use this flag.

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
```diff
@@ -8,30 +8,51 @@ import (
...
 	f := fs.Lookup("replace")
-	if f != nil {
-		if bp := generics.Cast[*bool](f.Value); bp != nil {
-			return
-		}
+	if f == nil {
+		fs.BoolVarP(&o.Replace, "replace", "R", false, "replace existing elements")
+	}
+
+	f = fs.Lookup("preserve-signature")
```
jakobmoellerdev, line 28: why do you use fs.Lookup before assigning variable? (in which case would this already be set?)

cmds/ocm/commands/ocmcmds/components/hash/cmd.go
```diff
@@ -180,17 +181,26 @@ func (h *action) Close() error {
...
+			for _, n := range h.norms {
+				err := h.write(h.mode.outfile, n)
+				if err != nil {
+					return err
+				}
+			}
+		} else {
+			dir := h.mode.outfile
+			dir = strings.TrimSuffix(dir, ".ncd")
+			err := h.ctx.FileSystem().Mkdir(dir, 0o755)
```
jakobmoellerdev, line 194: The filesystem permissions on this directory look too lax.

Wouldn't 0o644 suffice?

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
mandelsoft: the same optio set might be used multiple times, then the option can only be added once.

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
mandelsoft: there is a more detailed description as part of the command docu.

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
mandelsoft: there is a more detailed description as part of the command docu.

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
mandelsoft: done

api/helper/builder/ocm_identity.go
mandelsoft: it was intended as replacement for a sub structure like `b.ocm.meta`.

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
jakobmoellerdev: For me this does not replace the help for the option since that is what people will reference initially. @fabianburth wdyt?

cmds/ocm/commands/ocmcmds/common/addhdlrs/options.go
```diff
@@ -40,6 +70,9 @@ func (o *Options) Description() string {
 	return `
 The <code>--replace</code> option allows users to specify whether adding an
 element with the same name and extra identity but different version as an 
-existing element append (false) or replace (true) the existing element.
+existing element, append (false) or replace (true) the existing element.
+
+The <code>--preserve-signature</code> option prohibits changes of signature 
+relevant elements.
 `
```
mandelsoft, line 77: @jakobmoellerdev This is the docu for the options

jakobmoellerdev (approved):

PR #1028, opened 2024-10-30
chore: enhance the publishing to other repositories then github

hilmarf:

Publishing to external repositories might fail for various reasons, therefore we should be able to easily trigger the publishing to those repositories again and again without the need to rebuild a new release from scratch.


jakobmoellerdev (commented):
While I generally like the idea to have this separately publishable, this doesnt really solve the issue of making aware of the failure in the jobs. As such I would like to talk about reverting all of this behavior into regular jobs that can be retriggered via workflow_dispatch. Im happy to hear your thoughts though.

jakobmoellerdev (approved):
Will approve for now since it will be quite some time until we get to completely retouch this

PR #1033, opened 2024-10-31
fix artifact set tagging

mandelsoft:
Because the new OCI references now always contain the digest in addition to the tag, the artifactset creation
is not able anymore to provide appropriate meta data.

This is fixed by a new ArtVersion type able to handle the combination of digest and tag.


api/oci/ociutils/ref.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if v.Digest != nil {
+			vers += "@" + string(*v.Digest)
+		}
+		if vers == "" {
+			return "latest"
+		}
+	}
+	return vers
+}
```
jakobmoellerdev, line 64:
```suggestion
func (v *ArtVersion) VersionSpec() string {
	if v == nil {
	      return ""
	}
	
	var version string
      if v.Tag != nil {
		version = *v.Tag
	}
	if v.Digest != nil {
		version += "@" + string(*v.Digest)
	}
	// an initialized ArtVersion with nil fields is equivalent to latest
	if version == "" {
		version = "latest"
	}
	
	return version
}
```
just a nit, but easier to read due to only one layer of if

api/oci/ociutils/ref.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (v *ArtVersion) IsVersion() bool {
```
jakobmoellerdev, line 66: I think a method on a type called version with `IsVersion` is unintuitive.

```suggestion
func (v *ArtVersion) IsLatest() bool {
      return v != nil && v.Tag == nil && v.Digest == nil
}
```
then wherever you want to use IsVersion just use the inverse of IsLatest

jakobmoellerdev: I just noticed that all methods except for VersionSpec cannot deal with being nil. Is that intended?

api/oci/ociutils/ref.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"strings"
+
+	"github.com/mandelsoft/goutils/generics"
+	"github.com/opencontainers/go-digest"
+)
+
+// ParseVersion parses the version part of an OCI reference consisting
+// of an optional tag and/or digest.
+func ParseVersion(vers string) (ArtVersion, error) {
```
jakobmoellerdev, line 12: Because all methods on ArtVersion are defined on a pointer I believe this should become 
```suggestion
func ParseVersion(vers string) (*ArtVersion, error) {
```

cmds/ocm/commands/ocicmds/common/handlers/artifacthdlr/closure.go
```diff
@@ -52,7 +52,7 @@ func traverse(hist common.History, o *Object, octx out.Context) []output.Object
 					UniformRepositorySpec: o.Spec.UniformRepositorySpec,
 					ArtSpec: oci.ArtSpec{
 						Repository: o.Spec.Repository,
-						Digest:     &ref.Digest,
+						ArtVersion: oci.ArtVersion{Digest: &ref.Digest},
```
jakobmoellerdev, line 55: I think this version should be a pointer everywhere we reference it.

api/oci/ociutils/ref.go
```diff
@@ -0,0 +1,83 @@
...
+	if strings.HasPrefix(vers, "@") {
+		dig, err := digest.Parse(vers[1:])
+		if err != nil {
+			return ArtVersion{}, err
+		}
+		return ArtVersion{
+			Tag:    nil,
```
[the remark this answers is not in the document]
mandelsoft: removed

api/oci/ociutils/ref.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return ArtVersion{
+		Tag: &vers,
+	}, nil
+}
+
+// ArtVersion is the version part of an OCI reference consisting of an
+// optional tag and/or digest. Both parts may be nil, if a reference
+// does not include a version part.
+type ArtVersion struct {
```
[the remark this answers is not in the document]
mandelsoft: The reason is, that this is a type extracted from the already existing type oci.ArtSpec. It had to be moved out of the oci package to avoid a package usage cycle when using this extracted type.,

api/oci/ociutils/ref.go
mandelsoft: done

api/oci/ociutils/ref.go
mandelsoft: IsVersion is used as embedded type in ociArtSpec to be compatible here.

It returns whether this object (reference) (or nil)  specifies a dedicated version.

added check for nil

api/oci/ociutils/ref.go
mandelsoft: It the sibling for oci.ParseArt. It returned a value, therefore ParseVersion was done the same way. Unfortunately it has to be moved to another package to avoid cycles,

cmds/ocm/commands/ocicmds/common/handlers/artifacthdlr/closure.go
mandelsoft: see above

api/oci/ociutils/ref.go
jakobmoellerdev: Since ParseArt is also a ref creator, could we agree to move ParseArt to Pointer too? I assume you are talking about https://github.com/open-component-model/ocm/blob/69ea53a86ed3ae39e815f9bc73014bf49f3c731e/api/oci/ref.go#L227 and it seems like a pointer is applicable here too.

api/oci/ociutils/ref.go
jakobmoellerdev: Understood regarding the reuse. can we write a short comment on this? Then im totally fine with keeping it as is

cmds/ocm/commands/ocicmds/common/handlers/artifacthdlr/closure.go
jakobmoellerdev: What a mess :D

api/oci/ociutils/ref.go
mandelsoft: done

jakobmoellerdev (approved):

PR #1035, opened 2024-10-31
doc resource management

mandelsoft:
Extend the tour documentation to illustrate the rules for complying to the resource management used
by this library.


examples/lib/tour/docsrc/01-getting-started/README.md
```diff
@@ -216,3 +216,75 @@ sets the `X` flag for the written file.
...
+Many objects provided by the library offer some kind of resource management. In the example above, this is an
+OCM repository, the OCM component, component version and the access method.
+
+Those objects may use external resources, like temporary file system content or caches. To get rid of those resources again, they offer a `Close` method.
+
+To achieve the possibility to pass those objects around in non-functional call contexts they feature some kind of resource management. It allows to handle
+the life cycle of the resource in a completely local manner. To do so, a second method `Dup` is offered, which provides an independent reference to the original resources, which can be closed separately.
+The possible externally held resource are released with the close of the last reference.
+
+This offers a simple contract to handle resources in functions or object methods:
```
jakobmoellerdev, line 231: This examples reads very generic to me and is maybe not a good fit for the tour as part of getting started. Instead it is probably a good additional documentation to serve as context for the existing tour. WDYT?

mandelsoft: This is the place in the tour, where for the first time objects are provided featuring a `Close` methods.
Therefore, we should explain here, how a user should deal with such objects (even if Dup is not used, here.

For a regular straight-forward use case, there is typically no need to use Dup.

jakobmoellerdev: Is it the case that we want to give people an indepth explanation of how the resource management model works, even if they are just "getting started" in the CLI? I feel like it would deter people from moving on. If you still disagree, Im okay with a merge.

jakobmoellerdev: Addendum: Especially considering there is usually no need for Dup, maybe thats something not necessary in the guide?

mandelsoft: The reason is explicitly explained (case 3)

The section has been moved to a new last tour explaining the resource management and providing/describing a runnable code example.

examples/lib/tour/07-resource-management/example.go
```diff
@@ -0,0 +1,259 @@
...
+	err = semverutils.SortVersions(versions)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot sort versions")
+	}
+	fmt.Printf("versions for component ocm.software/ocmcli: %s\n", strings.Join(versions, ", "))
+
+	// Now, we have a look at the latest version. it is
+	// the last one in the list.
+	// --- begin lookup version ---
+	cv, err := c.LookupVersion(versions[len(versions)-1])
```
jakobmoellerdev, line 214: I have trouble accepting this code because it will require another make generate call on every release. Could we make this a static example so that it doesnt need to be updated constantly?

mandelsoft: I think, this would be another PR.

jakobmoellerdev (dismissed):
LGTM

jakobmoellerdev:
needs a regenerate due to my discussion issue

jakobmoellerdev (approved):

PR #1044, opened 2024-11-04
chore: reuse aggregation from ctf during component build

jakobmoellerdev:

this will create an aggregation post build that can be reused during releases if necessary

This is a revision of the component build in preparation to reuse the final CTF artifacts during our release


jakobmoellerdev:
Note to Reviewer: I have to change the PR rules due to the changed Job names after this goes in

hilmarf (approved):
LGTM - what a pity, that the aggregation again requires go-setup etc.

PR #1045, opened 2024-11-04
priority for CLI registration options

### LINKED ISSUE #1046: OCI artifact uploaders aren't used as expected
**What happened**:

When specifying an OCI artifact uploader on the command line it's not taken into account.

```bash
ocm transfer componentversions --copy-resources .\docker-hello-tiny ocm-oci-test.repository.foo.bar --uploader 'ocm/ociArtifacts={"ociRef":"ocm-oci-test.repository.foo.bar/images/foo"}'
```

Executing this results in something like this:

Image

But it was expected to have the image in the folder: `./images/foo/recmo/docker-hello-tiny`

Upload any docker image and specify an uploader like this: `--uploader 'ocm/ociArtifacts={"ociRef":"...`

There are three places where uploaders are specified:

1. Uploader defined on command line (should have highest priority by default)
2. Uploaders configured in `~/.ocmconfig` (should have higher priority than the builtin ones)
3. Internal ones - builtin (should be only used, when no other is specified)

mandelsoft:
The CLI registration options (--downloader, --uploader) now support an additional optional priority segment.
The new format is `<name>[:<artifact type>[:<media type>[:<prio>]]]=<JSON target config>`

The priority so far was the default priority also used for the builtin registrations. There competing registrations
from the CLI (or config) were never used.

Therefore
- the priority can now be specified on the command line
- the default priority is three times the default priority (300)
- the default priority for config objects is increased to twice the default priority (200)

Fixes #1046


jakobmoellerdev (commented):
Whats the concrete use case for this? I dont believe we have anyone asking for this (yet). Even if we do Im missing documentation on it (e.g. why would someone want to use it? and how is a priority specified, for example I could only know how it is used after looking into the test cases in code)

mandelsoft:
The problem was reported by @hilmarf . He wanted to register an OCI Uploader, which was never used because of its default priority.

cmds/ocm/commands/ocmcmds/common/options/optutils/registration.go
```diff
@@ -34,7 +44,7 @@ type RegistrationOption struct {
 	Registrations []*Registration
 }
 
-const RegistrationFormat = "<name>[:<artifact type>[:<media type>]]=<JSON target config"
+const RegistrationFormat = "<name>[:<artifact type>[:<media type>[:<prio>]]]=<JSON target config>"
```
hilmarf, line 47: Since it's user facing, can we please use complete words instead of abbreviations?

```suggestion
const RegistrationFormat = "<name>[:<artifact type>[:<media type>[:<priority>]]]=<JSON target config>"
```

hilmarf (approved):

PR #1057, opened 2024-11-07
bug: fix unmarshal consumer identity with empty value

fabianburth:
Fix empty identity value regression.

The regression was introduced after this [PR](https://github.com/open-component-model/ocm/pull/927) adding a custom unmarshaler for consumer identities which allows to specify values other than string (e.g. port: 5000), even though consumer identity is a map[string]string.

### DIFF
--- a/api/credentials/internal/identity.go
+++ b/api/credentials/internal/identity.go
@@ -93,6 +93,10 @@ func orMatcher(list []IdentityMatcher) IdentityMatcher {
 // ConsumerIdentity describes the identity of a credential consumer.
 type ConsumerIdentity map[string]string
 
+// UnmarshalJSON allows a yaml specification containing a data type other
+// string, e.g. a hostpath spec with a port. Previously, it would error if the
+// user specified `port: 5000` and instead, the user had to specify
+// `port: "5000"`.
 func (c *ConsumerIdentity) UnmarshalJSON(data []byte) error {
 	var m map[string]interface{}
 	err := runtime.DefaultJSONEncoding.Unmarshal(data, &m)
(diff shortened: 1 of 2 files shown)


api/ocm/extensions/accessmethods/ociartifact/method_test.go
```diff
@@ -55,6 +58,26 @@ var _ = Describe("Method", func() {
 			hint := spec.GetReferenceHint(&cpi.DummyComponentVersionAccess{env.OCMContext()})
 			Expect(hint).To(Equal("ocm/value:v2.0"))
 		})
+
+		FIt("", func() {
```
frewilhelm, line 62: Is `FIt` intended?

api/ocm/extensions/accessmethods/ociartifact/method_test.go
```diff
@@ -3,10 +3,13 @@ package ociartifact_test
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"ocm.software/ocm/api/credentials/extensions/repositories/directcreds"
 	. "ocm.software/ocm/api/helper/builder"
 	. "ocm.software/ocm/api/oci/testhelper"
+	"ocm.software/ocm/api/tech/oci/identity"
 
 	"ocm.software/ocm/api/credentials"
+	"ocm.software/ocm/api/credentials/config"
```
frewilhelm, line 12: I hate to nitpick this, but the import-order is not compliant :D

frewilhelm (commented):
Do you have any examples in which a consumer identity has an empty value and we don't want to throw an error or similar? It sounds odd to have a valid key but empty value^^

api/ocm/extensions/accessmethods/ociartifact/method_test.go
fabianburth: both shouldnt have been in the pr at all, sorry!

frewilhelm (approved):

Issue #995, opened 2024-10-21
Ensure that bigger (breaking) changes and potential breaking candidates are better and earlier communicated in the Release Notes


jakobmoellerdev:
**What would you like to be added**:

- Add BREAKING notes on all potential candidates for PRs and allow manual editing of release notes and include that manual edit as MANDATORY process for every release
- Ensure that this goes hand in hand with release candidate creation (separate issue, TBD)

**Why is this needed**:

hilmarf:
IMHO it's fine to generate the release notes as DRAFT, but publishing should be done only after HUMAN review and correction.

jakobmoellerdev:
We now notify of breaking changes correctly and we have them via separate category in the release notes.

PR #1059, opened 2024-11-07
chore: allow publishing to Brew via custom script

jakobmoellerdev:

Part of the release rework (https://github.com/open-component-model/ocm/issues/995) that will allow us to move away from goreleaser for actual release work. This fully replaces the goreleaser workflow of updating the brew TAP with a PR that creates a versioned link in a PR towards the TAP repository. This allows

- versioned brew installations like `brew install open-component-model/tap/ocm@0.17.0` instead of just the latest release
- Allows us to run the goreleaser process without ever publishing any artifact externally


.github/workflows/publish-to-other-than-github.yaml
```diff
@@ -19,6 +19,60 @@ on:
...
+      run: |
+        formula=$(go run scripts/hack/brew \
+            --version ${{ env.RELEASE_VERSION }} \
+            --template scripts/hack/brew/ocm_formula_template.rb.tpl \
+            --outputDirectory Formula)
+        mkdir -p Aliases
+        ln -s Formula/$(basename formula) Aliases/ocm
+        rm -r scripts
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v7
```
hilmarf, line 66: öhmm... maybe stupid question, but where is the PR actually created? Automatically on the latest checked-out repo? Or always on the one, where the action runs?

hilmarf (approved):
LGTM

jakobmoellerdev:
Not fully tested yet need to verify a PR creation

.github/workflows/publish-to-other-than-github.yaml
jakobmoellerdev: PR is created on the Tap Repository.

This is an example PR created from an internal testing repository: https://github.com/open-component-model/homebrew-tap/pull/3
The testing run (only accessible by org members): https://github.com/open-component-model/ocm-cicd-playground/actions/runs/11748205455/job/32731771674

a working tap repository can be found in https://github.com/jakobmoellerdev/homebrew-tap that follows this method and can be used for testing installations like

```
brew install jakobmoellerdev/tap/ocm # redirects via the alias to always the latest published
brew install jakobmoellerdev/tap/ocm@0.17.0 # versioned publish
```

hilmarf (approved):

hilmarf:
We'll need some follow up tasks for the brew-tap-repo. Branch-protection-rules? Codeowners file for automated review assignments...

jakobmoellerdev:
I will have to think about this some more, but technically these stories were always needed but we never got around to it right?

jakobmoellerdev:
I created a Ruleset for now at least in evaluation mode, will follow up with some more stories right after

hilmarf:
I've copy & pasted now the codeowners file from ocm, so that new PRs in brew-tap are not missed and automatically assigned to 'Maintainers'

PR #1070, opened 2024-11-11
skip digest for component constructors

mandelsoft:
This PR adds the option `--skip-digest-generation` to the `ocm add components` command.
It generally disables the digest generation for all added resources.
Additionally, the resource description may set `skipDigestGeneration` to `true` to disable the digest 
generation  for this particular resource.

Fixes https://github.com/open-component-model/ocm-project/issues/322

### DIFF
--- a/docs/reference/ocm_add_componentversions.md
+++ b/docs/reference/ocm_add_componentversions.md
@@ -27,8 +27,8 @@ componentversions, componentversion, cv, components, component, comps, comp, c
       --lookup stringArray        repository name or spec for closure lookup fallback
   -O, --output string             output file for dry-run
   -R, --replace                   replace existing elements
-  -S, --scheme string             schema version (default "v2")
   -s, --settings stringArray      settings file with variable settings (yaml)
+      --skip-digest-generation    skip digest creation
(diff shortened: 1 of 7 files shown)


docs/reference/ocm_add_componentversions.md
```diff
@@ -27,8 +27,8 @@ componentversions, componentversion, cv, components, component, comps, comp, c
       --lookup stringArray        repository name or spec for closure lookup fallback
   -O, --output string             output file for dry-run
   -R, --replace                   replace existing elements
-  -S, --scheme string             schema version (default "v2")
```
jakobmoellerdev, line 30: Why was this removed in this PR? It looks like a different fix.

jakobmoellerdev (commented):
Generally, LGTM but looks like another fix slipped into this one regarding the default scheme handling.

docs/reference/ocm_add_componentversions.md
mandelsoft: because the AddFlags was missing....

jakobmoellerdev (approved):

frewilhelm (approved):

PR #1071, opened 2024-11-11
chore: release branches as `releases/vX.Y` instead of `releases/vX.Y.Z`

jakobmoellerdev:

This reworks the release branch action to correctly branch off in patterns that do not include Z / Patch versions:

```
releases/v0.17.0 # before
releases/v0.17 # after
```

To achieve this it allows to print only the major/minor combination for the release_generate.go generator.

Additionally, it adds an automatic bump to main after the Branch Cutoff was completed into main.

Currently our release branches include the 0 Z / Patch version as part of their name, but this is misleading because we also use these branches for patches onto the Y / Minor version.

This fixes up the release branch creation to

1. Create the Branch according to this pattern
2. Fixup the Main Branch with a Version Bump PR after the bump so that the next minor development can start.

part of https://github.com/open-component-model/ocm/issues/995 to ensure proper candidate creation


jakobmoellerdev:
Example for a Run of this PR in the internal member-only testing Repository:

https://github.com/open-component-model/ocm-cicd-playground/actions/runs/11721978049

(The run also included the automatic creation of a first release candidate but this is left out of this PR)

.github/workflows/release-branch.yaml
```diff
@@ -1,75 +1,136 @@
...
+          set -e
+          minor="$(go run ./api/version/generate print-major-minor)"
+          echo "minor=$minor" >> $GITHUB_OUTPUT
+          echo "Current Major-Minor Version: $minor"
+      - name: Verify Branch does not exist
+        id: verify-branch
+        run: |
+          set -e
+          minor="v${{ steps.get-minor.outputs.minor }}"
+          branch="releases/$minor"
```
hilmarf, line 55: is it releases/x.y.z or release/x.y.z ?

jakobmoellerdev: it is `releases` (plural). See https://github.com/open-component-model/ocm/branches/all?query=releases

.github/workflows/release-branch.yaml
```diff
@@ -1,75 +1,136 @@
...
-          exit 1
-        fi
-        echo "creating release branch for $tag"
-        n="releases/$tag"
-        git checkout -b "$n"
-        v="$(go run ./api/version/generate bump-patch)"
-        echo "$v" > VERSION
-        git add VERSION
-        git commit -m "Prepare Development of v$v"
-        git push origin "$n"
```
hilmarf, line 75: so version correction on release branch happens later, right?

hilmarf (approved):

.github/workflows/release-branch.yaml
jakobmoellerdev: since main will be on the correct version during cutoff that is not needed.

Example:

1. Main is on 0.17.0-dev
2. Now we want to prepare the first candidate for 0.17.0 (rc.1), Thus we trigger this cutoff action
3. In this case, the branch now does not need an additional commit because 0.17.0-dev is correct
4. Main then gets bumped via PR to 0.18.0-dev
5. We can now build on Main (as 0.18.0-dev candidates) or on the release branch (as 0.17.0-dev candidates) or trigger the release (on 0.17.0-dev this then leads to 0.17.0-rc1)

PR #1076, opened 2024-11-11
chore: cleanup release action

jakobmoellerdev:

This fixes several inconsistencies with the release action and cleans it up before we are migrating some of the more substantial work within it:

1. main no longer gets release notes (as only release branches need to hold release notes)
2. The release version fetching and resolution is separated into a job just for this purpose, and can be reused across steps that need access to it
3. The draft release notes are now fetched via `gh` instead of the old `cardinalby/git-get-release-action@v1` because it is much faster and transparent what happens and the action wasn't maintained properly.
4. The `prerelease` flag was renamed to `release_candidate_name` to make it more clear for what it is actually used in the code. The existing docs on the action one can see when triggering them are unchanged
5. The `create_branch` option is killed because we dont expect the release to trigger the release branch creation, we expect that to come before. A separate PR can trigger the first release candidate creation from the Branch cutoff however (this will come separately)

Part of the release process and transparency rework in https://github.com/open-component-model/ocm/issues/995


jakobmoellerdev:
A partial run of the changes can be observed here: https://github.com/open-component-model/ocm-cicd-playground/actions/runs/11783730801 (cancelled before publish to avoid artifacts being pushed)

.github/workflows/release.yaml
```diff
@@ -8,73 +8,76 @@ on:
...
       env:
-        GITHUB_TOKEN: ${{ github.token }}
-      with:
-        draft: true
-        releaseName: ${{ env.BASE_VERSION }}
-
+        GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+      run: |
+        RELEASE_JSON=$( \
+          gh api /repos/open-component-model/ocm-cicd-playground/releases \
```
hilmarf, line 72:
```suggestion
          gh api /repos/open-component-model/ocm/releases \
```

hilmarf (dismissed):
besides the `cicd-playground` looks good

.github/workflows/release.yaml
jakobmoellerdev: changed to github repository variable

hilmarf (approved):

PR #1087, opened 2024-11-14
feat: migrate to fully distroless ocm image

jakobmoellerdev:

Migrates to Distroless which has a few advantages over alpine:

1. Smaller Base Image
2. No Package Manager or Shell that could be used to exploit vulnerabilities
3. Non-Root user by default

I also remove the latest check in alpine since it only fetched a tag anyhow and we should pin by digest


jakobmoellerdev:
In addition to this PR, longterm I want to completely get rid of the component based dockerfile and switch to a central one so that everyone can use docker build instead of 2 dockerfiles that differ in the build process. But I want to split this into a separate one, so for now, this should suffice

Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG GO_VERSION="1.23"
 ARG ALPINE_VERSION="3.20"
+ARG DISTROLESS_VERSION=debian12@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
```
frewilhelm, line 3: i guess thats a version pin. Do we have a bot that checks for updates?

Dockerfile
```diff
@@ -20,26 +21,21 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
-LABEL org.opencontainers.image.description="Open Component Model command line interface based on Alpine ${ALPINE_VERSION}"
+LABEL org.opencontainers.image.description="Open Component Model command line interface based on Distroless ${DISTROLESS_VERSION}"
 LABEL org.opencontainers.image.vendor="SAP SE"
 LABEL org.opencontainers.image.licenses="Apache-2.0"
 LABEL org.opencontainers.image.url="https://ocm.software/"
 LABEL org.opencontainers.image.source="https://github.com/open-component-model/ocm"
 LABEL org.opencontainers.image.title="ocm"
 LABEL org.opencontainers.image.documentation="https://github.com/open-component-model/ocm/blob/main/docs/reference/ocm.md"
-LABEL org.opencontainers.image.base.name="alpine:${ALPINE_VERSION}"
+LABEL org.opencontainers.image.base.name="gcr.io/distroless/static-${DISTROLESS_VERSION}"
```
frewilhelm, line 38: why do we use this domain `org.opencontainers.image.base.name`?

components/ocmcli/Dockerfile
```diff
@@ -1,21 +1,18 @@
 # This Dockerfile is used by `make` via the Makefile
 
-ARG ALPINE_VERSION="3.20.2"
-FROM --platform=$BUILDPLATFORM alpine:${ALPINE_VERSION}
-
-# Create group and user
-ARG UID=1000
-ARG GID=1000
-RUN addgroup -g "${GID}" ocmGroup && adduser -u "${UID}" ocmUser -G ocmGroup -D
+# debian12:nonroot
```
frewilhelm, line 3: I guess thats the default/only user of that image? I would expect an explanation for this comment or delete it

Dockerfile
jakobmoellerdev: No we do not, yet :D . I can create an issue, thanks

components/ocmcli/Dockerfile
jakobmoellerdev: The "nonroot" images are configured to default to executing as a normal user account called "nonroot", with user ID 65532. That is the only difference. All the images contain configuration for the user accounts root, nobody, and nonroot.

frewilhelm: But do we need the comment?

jakobmoellerdev: Explicitly added it via the image to avoid needing the comment. Please double check

frewilhelm:
> In addition to this PR, longterm I want to completely get rid of the component based dockerfile and switch to a central one so that everyone can use docker build instead of 2 dockerfiles that differ in the build process. But I want to split this into a separate one, so for now, this should suffice

frewilhelm:
Please adjust `COPY go.mod go.sum *.go VERSION ./` to `COPY go.mod go.sum ./` in `./Dockerfile`

frewilhelm (commented):
Overall the changes lgtm. However, we should discuss `component/ocmcli/Makefile` and its `Dockerfile` (as you already proposed).

frewilhelm (approved):

jakobmoellerdev:
not to be merged until @hilmarf explicitly approves due to cross dependency to piper step

PR #1090, opened 2024-11-15
chore: make sure that version bumping happens everytime

jakobmoellerdev:

This makes sure that we dont only do a patch bump on main after a successful branch cut, but we also do a bump on the release branch (for the next z / patch version) after the minor has been released.

Creates a shared workflow that can be called from other workflows.

Ensures more automated workflows after the rework from https://github.com/open-component-model/ocm/issues/995

Example run at https://github.com/open-component-model/ocm-cicd-playground/actions/runs/11838213132/job/32986884073


.github/workflows/release.yaml
```diff
@@ -194,7 +194,7 @@ jobs:
       run: |
         set -e
         git checkout ${GITHUB_REF#refs/heads/}
-        v="$(go run ./api/version/generate bump-version)"
+        v="$(go run ./api/version/generate bump-minor)"
```
hilmarf, line 197: hmm... what happens when we need to create a patch-release (e.g. `v0.17.1`)? Then this is executed on `releases/vX.Y` branch right? Or am I wrong? Wouldn't then the version be bumped to `v0.18.x` ?

hilmarf (dismissed):

frewilhelm (dismissed):

frewilhelm (approved):
lgtm (even though I have to admit that this whole generate-version scripts/programs are really confusing)

jakobmoellerdev:
Understood. If you could let me know how we can improve them, please tell me straight away. I will try to simplify as much as possible but currently my PRs are focusing on only "touching up" without rewriting

frewilhelm:
I created an issue https://github.com/open-component-model/ocm/issues/1094. Although I agree, that this has no priority.

PR #1098, opened 2024-11-18
chore: also create a tag to avoid dangling commits

jakobmoellerdev:

currently our tagged releases do not have branch accomodating them, leading to dangling commits. (see https://github.com/open-component-model/ocm/tree/v0.18.0-rc.1 for an example)

This fixes this by also creating an appropriate release branch.

This might introduce issues with things like the release drafter though, so this is likely only the first line of commits for this.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -160,7 +160,8 @@ jobs:
         git commit -m "Release $RELEASE_VERSION"
         msg="Release ${{ env.RELEASE_VERSION }}"
         git tag --annotate --message "${msg}" ${{ env.RELEASE_VERSION }}
-        git push origin ${{ env.RELEASE_VERSION }}
+        # push both the tag as well as a release branch with that tag.
+        git push origin releases/${{ env.RELEASE_VERSION }} ${{ env.RELEASE_VERSION }}
 
     - name: Create GPG Token file from Secret
       run: |


Skarlso (approved):

.github/workflows/release.yaml
```diff
@@ -160,7 +160,8 @@ jobs:
         git commit -m "Release $RELEASE_VERSION"
         msg="Release ${{ env.RELEASE_VERSION }}"
         git tag --annotate --message "${msg}" ${{ env.RELEASE_VERSION }}
-        git push origin ${{ env.RELEASE_VERSION }}
+        # push both the tag as well as a release branch with that tag.
+        git push origin releases/${{ env.RELEASE_VERSION }} ${{ env.RELEASE_VERSION }}
```
Skarlso, line 164: Hmmm... The release notes will not be on this new tag that is created above.

Skarlso (changes requested):
The release note is only pushed into the main branch.

.github/workflows/release.yaml
Skarlso: Or rather... whatever GITHUB_REF is... maybe it's fine?

Skarlso (approved):

.github/workflows/release-drafter.yaml
```diff
@@ -3,7 +3,10 @@ name: Release Drafter
 on:
   push:
     branches:
-      - releases/*
+      # we only want to draft release notes for the base version,
+      # e.g. for releases/v0.18, we draft v0.18.0 for all RCs
+      # this means that we explicitly don't want to draft release notes for releases/v0.18.0 or releases/v0.18.1
+      - 'releases/v[0-9]+.[0-9]+'
```
Skarlso, line 9: This matches `releases/v10.0.3`. Use `releases/v[0-9]+.[0-9]+$`

jakobmoellerdev: because the last `[0-9]+` is not a `.` this expression should also work here

Skarlso: Huh 😱 Okay. :D

Skarlso (approved):

PR #1100, opened 2024-11-19
chore: add correct labels for flake nix job

jakobmoellerdev:

makes sure they disappear from the release notes since they are just bumps in main that people dont really care about

### DIFF
--- a/.github/workflows/flake_vendorhash.yaml
+++ b/.github/workflows/flake_vendorhash.yaml
@@ -57,5 +57,8 @@ jobs:
           branch: nix/flake
           delete-branch: true
           sign-commits: true
+          labels: |
+            kind/skip-release-notes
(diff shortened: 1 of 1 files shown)


frewilhelm (approved):

.github/workflows/flake_vendorhash.yaml
```diff
@@ -57,5 +57,8 @@ jobs:
           branch: nix/flake
           delete-branch: true
           sign-commits: true
+          labels: |
+            kind/skip-release-notes
+            chore
```
hilmarf, line 62: which one will then be taken?

jakobmoellerdev: both but since skip is mentioned it will be skipped completely

PR #1103, opened 2024-11-19
chore: rework labeling jobs

jakobmoellerdev:

this unifies the labelling jobs in one workflow and makes sure that at least one of our identifying labels is always set


.github/workflows/pr-labeler-kind.yaml
```diff
@@ -16,4 +20,39 @@ jobs:
...
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          xs_label: 'size/xs'
+          xs_max_size: '10'
+          s_label: 'size/s'
+          s_max_size: '100'
+          m_label: 'size/m'
+          m_max_size: '500'
+          l_label: 'size/l'
+          l_max_size: '10000'
```
frewilhelm, line 41: I know `10000` LoC is our current maximum, but should we talk again about that threshold as it is quite large?

jakobmoellerdev: Yes, but please lets not do this in this PR. Would you mind creating a separate issue for this?

jakobmoellerdev: https://github.com/open-component-model/ocm/issues/1107

frewilhelm (approved):

PR #1108, opened 2024-11-19
chore: ensure that PR titles must be semantic

jakobmoellerdev:

This enforces Semantic Pull Request titles:
- see https://github.com/amannn/action-semantic-pull-request?tab=readme-ov-file#examples for examples
- see https://www.conventionalcommits.org/en/v1.0.0/#specification for the full specification

We want this because we use squashing so all squashed commits should conform to our commit style

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,6 +1,12 @@
-name: PR Labeler
+name: Pull Request
 on:
-- pull_request_target
+  pull_request_target:
+    # make sure that when the PR changes, we also update
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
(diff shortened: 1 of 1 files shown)


jakobmoellerdev:
Sample run of title failure: https://github.com/open-component-model/ocm-cicd-playground/actions/runs/11911808386/job/33194099938

jakobmoellerdev:
https://github.com/open-component-model/ocm/actions/runs/11913289589/job/33198678380?pr=1108 fails because there was a typo in the label config and it checks for `hore` instead of `chore` this must be forcefully overwritten right now

hilmarf (approved):

Skarlso (approved):

PR #1113, opened 2024-11-20
docs: document complex artifact transfer

jakobmoellerdev:

This documents a more complicated scenario for an artifact transfer that might be useful if only a single OCI image needs to be copied to a different repository or registry in OCI.


frewilhelm (commented):
lgtm - but I am pretty sure this change will require a `make generate` or the like

hilmarf (approved):

PR #1115, opened 2024-11-20
chore: adjust labels to new configuration

hilmarf:

When we migrate our labels and use [only the central managed ones](https://github.com/open-component-model/.github/blob/main/.github/settings.yml#L62), we'll need to change them at several places.

Once we've merged this, we should execute the migration script once. And then https://github.com/open-component-model/ocm/pull/1150


.github/workflows/pull-request.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: PRs should have at least one qualifying label
         uses: docker://agilepathway/pull-request-label-checker:latest
         with:
-          one_of: chore,fix,bugfix,bug,kind/bug,feature,enhancement,kind/enhancement,dependencies
+          one_of: kind/chore,kind/fix,kind/feature,kind/dependency,kind/documentation,kind/refactor,kind/skip-release-notes,kind/epic,kind/poc
```
jakobmoellerdev, line 64: why is it acceptable to have a PR with kind/epic or kind/poc? IMO those should never be merged. WDYT?

.github/release-drafter.yml
```diff
@@ -9,30 +9,35 @@ template: |
...
 exclude-labels:
   - 'kind/skip-release-notes'
-  - 'wontfix'
-  - 'triage/wont-fix'
-  - 'triage/invalid'
+  - 'dev/wont-fix'
+  - 'component/github-actions'
 categories:
-  - title: '‼️ Breaking Changes'
+  - title: '☠ Breaking Changes ☠'
```
jakobmoellerdev, line 15: I fear that you might be taking the emoji game a bit far :D

.github/workflows/pull-request.yaml
hilmarf: yes and no... when a PR closes an epic issue, which label should be used then for the PR?
e.g. "replacement of docker with oras" could be refactoring or epic? Or major API restructuring...

.github/release-drafter.yml
fabianburth: I am all for the emoji game, but I'd agree the skull might go a bit too far ;) 
How about a warning or stop sign? '⚠️ Breaking Changes' / '🚫 Breaking Changes'? 

But if you are a big fan of the skull, don't let me hold you back!

morri-son: I prefer the Warning sign with the exclamation mark. This is also what Gardener has chosen:

:warnung: Breaking Changes

https://gardener-cloud.slack.com/archives/C041L0SLZ09/p1731573289318969

.github/workflows/pull-request.yaml
jakobmoellerdev: IMO a PR that closes an epic doesnt need a kind/epic level, especially because there should never be a single PR that closes an entire epic.

.github/release-drafter.yml
hilmarf: ⚠️ Breaking Changes

.github/config/labeler.yml
```diff
@@ -1,9 +1,27 @@
 # see https://github.com/actions/labeler?tab=readme-ov-file#match-object to configure correctly
-dependencies:
+kind/dependency:
```
jakobmoellerdev, line 2: very cool!

jakobmoellerdev (commented):
I am missing adjustments to https://github.com/open-component-model/ocm/pull/1115/files#diff-a8619735ff14304aa0514284f86ff5145b0a6bae2e76a37faeb0ad899a3d8db4R25 to accomodate the label changes.

hilmarf:
> I am missing adjustments to https://github.com/open-component-model/ocm/pull/1115/files#diff-a8619735ff14304aa0514284f86ff5145b0a6bae2e76a37faeb0ad899a3d8db4R25 to accomodate the label changes.

here it is:
https://github.com/open-component-model/ocm/pull/1115/files#diff-a8619735ff14304aa0514284f86ff5145b0a6bae2e76a37faeb0ad899a3d8db4R25-R39

hilmarf:
blocked by: https://github.com/open-component-model/.github/pull/14

.github/workflows/pull-request.yaml
```diff
@@ -24,18 +24,18 @@ jobs:
...
-              "feat":"kind/enhancement",
-              "fix":"fix",
-              "chore":"chore",
-              "docs":"kind/documentation",
-              "test":"kind/test",
-              "perf":"kind/performance"
+              "feat":"kind/feature",
+              "fix":"kind/bugfix",
+              "chore":"kind/chore",
+              "docs":"area/documentation",
```
jakobmoellerdev, line 30: qq: why is it area documentation or area/testing instead of kind?

hilmarf: we had both in the past: `area/` and `kind/` - since the labels are more used by PO on issues, than developers on PRs, I thought, I'll keep only the `area/`

jakobmoellerdev: I think this might be something to revisit because any documentation we update in OCM (e.g. examples) would be labelled as such. Is this okay?

hilmarf: > I think this might be something to revisit because any documentation we update in OCM (e.g. examples) would be labelled as such. Is this okay?

Why not? If we plan to get some assistance from an information-engineer to revisit all our documentation for example. But if you have concerns, we can remove that.

jakobmoellerdev: I was expecting all PRs to have `kind` instead of `area` because 
- an `area` is where something belongs, while 
- `kind` is a definition of what something is

We are now unifying things into different prefixes and I really dont like it. But lets consider this as nit and not as blocking comment

jakobmoellerdev (approved):

PR #1118, opened 2024-11-20
feat(log): log http requests for OCI and docker based on trace level by injecting a logger

jakobmoellerdev:

it is now possible to inject a trace attribute to OCMs logging architecture to allow tracing back HTTP calls for oci registries and the docker client:

```
ocm --logkeys /+ocm/oci=trace [YOUR COMMAND INTERACTING WITH OCI HERE]
```

Note that this does not take care of all access types yet because we dont have a unified http.Client.

Note that it is now also possible to pass `/+ocm/docker=trace` to enable logging for the docker client infrastructure, or set `--loglevel=trace` to get a full tracelog with HTTP statements.

Authorization Headers are redacted

### DIFF
--- a/api/oci/extensions/repositories/ocireg/repository.go
+++ b/api/oci/extensions/repositories/ocireg/repository.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"net/http"
 	"path"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


api/utils/logging/roundtripper.go
```diff
@@ -0,0 +1,36 @@
...
+	logger logging.Logger
+	http.RoundTripper
+}
+
+func (t *RoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Redact the Authorization header to make sure it doesn't get logged at any point.
+	header := req.Header
+	if key := "Authorization"; req.Header.Get(key) != "" {
+		header = header.Clone()
+		header.Set(key, "***")
```
ccwienk, line 27: nit: might print first/last two or three characters / `Bearer ` + first few characters (may be helpful in certain cases of debugging).

jakobmoellerdev: Thought about this but then remembered that basic auth has a different Header and there are cases where basic auth might still be used.

Dont really want to do a string parsing on that header to either get `Basic` or `Bearer` out, so gonna omit for now.

fabianburth (approved):
Looks good

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -112,6 +114,20 @@ func (r *RepositoryImpl) getCreds(comp string) (credentials.Credentials, error)
...
+
+func (t *LoggingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	t.logger.Trace("roundtrip",
+		"url", req.URL,
+		"method", req.Method,
+		"header", req.Header,
+	)
+	return t.Transport.RoundTrip(req)
+}
+
```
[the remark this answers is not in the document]
jakobmoellerdev: done

PR #1123, opened 2024-11-21
docs: finally some working examples for a lot of commands

jakobmoellerdev:

This adds various indirections on the OCM CLIs own component versions as examples instead of kubelink, which never worked.

fix https://github.com/open-component-model/ocm-project/issues/20


Skarlso (approved):
Well done! 👏

hilmarf (approved):
Nice!

PR #1077, opened 2024-11-12
chore: setup release to reuse CTF from components workflow

jakobmoellerdev:
#### What this PR does / why we need it

This makes sure that the CTF is only built once in the release flow (during the components build) and then reused across the release. For this it needs to be built with the correct version during release (e.g. v0.18.0-rc.1) which we now pass via `EFFECTIVE_VERSION` into the Makefiles in the repo.

This allows us to get rid of the `make CTF_TYPE=directory ctf` step in the goreleaser configuration so we now no longer need to build the ctf twice.

#### Which issue(s) this PR fixes

Part of the release candidate transparency rework in https://github.com/open-component-model/ocm/issues/995 as it allows us to prepare the release to only use the binaries from the CTFs.

PR #1124, opened 2024-11-22
chore: fixup release action versioning and notes process

jakobmoellerdev:

Makes sure that release notes are correctly drawn in and that we always have the version set correctly in our CTFs.

This fixes up the release action in 2 important ways:

1. As part of the removal of the release notes we now no longer have a file to use for release notes to give to goreleaser. This means we have to temporarily create one
2. While discovering this I noticed that the Component Versions that are built, are actually using the `VERSION` variable while `EFFECTIVE_VERSION` is a combination of VERSION+COMMIT by default. I didnt want to change this behavior originally, so now I started to use VERSION for the Makefiles (which also means I had to export them as settable). This now leads to correct CTF versions `v0.19.0-rc.1` instead of `v0.19.0-dev` for example.

As such its a follow up from https://github.com/open-component-model/ocm/pull/1120 and https://github.com/open-component-model/ocm/pull/1077

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,52 +38,52 @@ jobs:
     outputs:
       draft-release-notes: ${{ steps.release-notes.outputs.json }}
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - name: Check Tag
-        run: |
-          set -e
-          if git ls-remote --exit-code origin refs/tags/${{ env.RELEASE_VERSION }} ; then
-            >&2 echo "tag ${{ env.RELEASE_VERSION }} already exists"
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Check Tag
+      run: |
+        set -e
+        if git ls-remote --exit-code origin refs/tags/${{ env.RELEASE_VERSION }} ; then
+          >&2 echo "tag ${{ env.RELEASE_VERSION }} already exists"
+          exit 1
+        fi
+    - name: Check if release is running on release branch
+      run: |
+        if [[ ${{ env.REF }} != *"releases/"* ]]; then
+            echo "The branch ${{ env.REF }} is not a valid release branch and cannot be used for a release"
             exit 1
-          fi
-      - name: Check if release is running on release branch
-        run: |
-          if [[ ${{ env.REF }} != *"releases/"* ]]; then
-              echo "The branch ${{ env.REF }} is not a valid release branch and cannot be used for a release"
-              exit 1
-          fi
-          echo "Branch ${{ env.REF }} is a valid release branch"
-      - name: Generate token
-        id: generate_token
-        uses: tibdex/github-app-token@v2
-        with:
-          app_id: ${{ secrets.OCMBOT_APP_ID }}
-          private_key: ${{ secrets.OCMBOT_PRIV_KEY }}
-      - name: Ensure existing Draft Release Notes exist
-        id: release-notes
-        shell: bash
-        env:
-          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
-        run: |
-          RELEASE_JSON=$( \
-            gh api /repos/${{ github.repository }}/releases \
-            -q '.[] | select(.name == "${{ env.RELEASE_VERSION_NO_SUFFIX }}" and .draft == true)' \
-          )
-          echo "json=${RELEASE_JSON}" >> $GITHUB_OUTPUT
-          # if no draft release notes are found, we cannot continue
-          if [ -z "${RELEASE_JSON}" ]; then
-            echo "No draft release notes found for ${{ env.RELEASE_VERSION_NO_SUFFIX }}"
-            exit 1
-          fi
+        fi
+        echo "Branch ${{ env.REF }} is a valid release branch"
+    - name: Generate token
+      id: generate_token
+      uses: tibdex/github-app-token@v2
+      with:
+        app_id: ${{ secrets.OCMBOT_APP_ID }}
+        private_key: ${{ secrets.OCMBOT_PRIV_KEY }}
+    - name: Ensure existing Draft Release Notes exist
+      id: release-notes
+      shell: bash
+      env:
+        GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+      run: |
+        RELEASE_JSON=$( \
+          gh api /repos/${{ github.repository }}/releases \
+          -q '.[] | select(.name == "${{ env.RELEASE_VERSION_NO_SUFFIX }}" and .draft == true)' \
+        )
+        echo "json=${RELEASE_JSON}" >> $GITHUB_OUTPUT
+        # if no draft release notes are found, we cannot continue
+        if [ -z "${RELEASE_JSON}" ]; then
+          echo "No draft release notes found for ${{ env.RELEASE_VERSION_NO_SUFFIX }}"
+          exit 1
+        fi
   components:
     name: Component CTF Builds
     uses: ./.github/workflows/components.yaml
     needs: [ check,release-version ]
     with:
-      effective-version: ${{ needs.release-version.outputs.version_no_prefix }}
+      version: ${{ needs.release-version.outputs.version_no_prefix }}
       upload-ctf: true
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yaml
```diff
@@ -183,19 +182,24 @@ jobs:
...
     - name: Create GPG Token file from Secret
       run: |
         echo "${{ secrets.GPG_PRIVATE_KEY_FOR_SIGNING }}" > ocm-releases-key.gpg
         echo "GPG_KEY_PATH=ocm-releases-key.gpg" >> $GITHUB_ENV
 
+    - name: Setup Release Notes as File to make it readable by GoReleaser
+      run: |
+        echo "${{ fromJSON(needs.check.outputs.draft-release-notes).body }}" > $RUNNER_TEMP/RELEASE_NOTES.md
+        echo "RELEASE_NOTES=$RUNNER_TEMP/RELEASE_NOTES.md" >> $GITHUB_ENV
+        cat $RUNNER_TEMP/RELEASE_NOTES.md
```
hilmarf, line 196: debug leftover or intention?

components/helmdemo/Makefile
```diff
@@ -8,9 +8,11 @@ CTF_TYPE ?= directory
 HELMINSTCOMP = $(PROVIDER)/toi/installers/helminstaller
 
 REPO_ROOT                                     := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
-VERSION                                        = $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+ifeq ($(VERSION),)
+VERSION                                        := $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+endif
 COMMIT                                         = $(shell git rev-parse --verify HEAD)
-# if EFFECTIVE_VERSION is not set, set it to VERSION-COMMIT
+# if EFFECTIVE_VERSION is not set, VERSION+COMMIT
```
hilmarf, line 15:
```suggestion
# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMIT
```

components/helminstaller/Makefile
```diff
@@ -9,9 +9,11 @@ IMAGE_PLATFORMS ?= linux/amd64 linux/arm64
 CTF_TYPE ?= directory
 
 REPO_ROOT                                      := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
+ifeq ($(VERSION),)
 VERSION                                        := $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+endif
 COMMIT                                         := $(shell git rev-parse --verify HEAD)
-# if EFFECTIVE_VERSION is not set, set it to VERSION-COMMIT
+# if EFFECTIVE_VERSION is not set, VERSION+COMMIT
```
hilmarf, line 16:
```suggestion
# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMIT
```

hilmarf (approved):
you're planning to get rid of the individual makefiles sooner or later, right?
Otherwise it might be worth thinking of using [include ...](https://web.mit.edu/gnu/doc/html/make_3.html#SEC15) ?

components/demoplugin/Makefile
```diff
@@ -7,7 +7,9 @@ PLATFORMS ?= linux/amd64 linux/arm64
 CTF_TYPE ?= directory
 
 REPO_ROOT                                      := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
-VERSION                                        = $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+ifeq ($(VERSION),)
```
frewilhelm, line 10: I am not that in to `Makefile` syntax, but I would guess that `Version ?= $(...)` should produce the same result as the condition

frewilhelm: See `6.2.4 Conditional Variable Assignment` in https://www.gnu.org/software/make/manual/make.html

.github/workflows/release.yaml
jakobmoellerdev: no i wanted to have the real release notes logged in the stepped. do you think thats a problem?

components/demoplugin/Makefile
jakobmoellerdev: https://github.com/open-component-model/ocm/pull/1124/files#diff-b0b6d9a4b18557276003816e5b6516ee6e07edbd1bf823505e9127530ec8917dR14-R15

jakobmoellerdev: #trustmeitried

jakobmoellerdev:
> you're planning to get rid of the individual makefiles sooner or later, right?
Otherwise it might be worth thinking of using [include ...](https://web.mit.edu/gnu/doc/html/make_3.html#SEC15) ?

Currently unplanned, but we can create an issue for this so I can take care of it. I would really like to just get the releases working right now.

components/demoplugin/Makefile
jakobmoellerdev: From 6.2.4

> Note that a variable set to an empty value is still defined, so ‘?=’ will not set that variable.

components/helmdemo/Makefile
```diff
@@ -8,9 +8,11 @@ CTF_TYPE ?= directory
 HELMINSTCOMP = $(PROVIDER)/toi/installers/helminstaller
 
 REPO_ROOT                                     := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
-VERSION                                        = $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+ifeq ($(VERSION),)
+VERSION                                        := $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+endif
 COMMIT                                         = $(shell git rev-parse --verify HEAD)
-# if EFFECTIVE_VERSION is not set, set it to VERSION-COMMIT
+# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMITversion
```
hilmarf, line 15:
```suggestion
# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMIT
```

components/helminstaller/Makefile
```diff
@@ -9,9 +9,11 @@ IMAGE_PLATFORMS ?= linux/amd64 linux/arm64
 CTF_TYPE ?= directory
 
 REPO_ROOT                                      := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
+ifeq ($(VERSION),)
 VERSION                                        := $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+endif
 COMMIT                                         := $(shell git rev-parse --verify HEAD)
-# if EFFECTIVE_VERSION is not set, set it to VERSION-COMMIT
+# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMITversion
```
hilmarf, line 16:
```suggestion
# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMIT
```

components/subchartsdemo/Makefile
```diff
@@ -13,9 +13,11 @@ PODINFO_CHART_VERSION = 6.3.5
 
 REPO_ROOT                                     := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))/../..
 GIT_TREE_STATE                                 = $(shell [ -z "$$(git status --porcelain 2>/dev/null)" ] && echo clean || echo dirty)
-VERSION                                        = $(shell go run $(REPO_ROOT)/api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+ifeq ($(VERSION),)
+VERSION                                        := $(shell go run ../../api/version/generate/release_generate.go print-rc-version $(CANDIDATE))
+endif
 COMMIT                                         = $(shell git rev-parse --verify HEAD)
-# if EFFECTIVE_VERSION is not set, set it to VERSION-COMMIT
+# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMITversion
```
hilmarf, line 20:
```suggestion
# if EFFECTIVE_VERSION is not set, set it to VERSION+COMMIT
```

hilmarf (approved):

hilmarf (approved):

PR #1137, opened 2024-11-25
chore(github_actions): using now classic secret of OCM_CI_ROBOT

hilmarf:

see: https://github.com/open-component-model/ocm/pull/1133

### DIFF
--- a/.github/workflows/publish-to-other-than-github.yaml
+++ b/.github/workflows/publish-to-other-than-github.yaml
@@ -130,24 +130,11 @@ jobs:
       run: |
         $version = "${{ github.event.client_payload.version }}" -replace 'v'
         echo "RELEASE_VERSION=$version" | Out-File $env:GITHUB_ENV
-    - name: Generate token
-      id: generate_token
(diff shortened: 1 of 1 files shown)


frewilhelm (approved):

.github/workflows/publish-to-other-than-github.yaml
```diff
@@ -130,24 +130,11 @@ jobs:
...
-    - name: Checkout
-      uses: actions/checkout@v4
-      with:
-        token: ${{ steps.generate_token.outputs.token }}
     - name: Install wingetcreate
-      run: choco install wingetcreate
+      run: iwr https://aka.ms/wingetcreate/latest -OutFile wingetcreate.exe
     - name: Update Winget package
       run: |
-        cd hack\winget
```
frewilhelm, line 149: do we need that repository?

hilmarf: actually I'm not 100% sure, it was necessary to create the very first version, but I hope the update is working without it... then we can also remove the complete folder.

PR #1139, opened 2024-11-25
chore: rework release note handling

jakobmoellerdev:

reworks release note management based on github release note api

It does this in the following way:
1. Creates a tag on cutoff that is annotated
2. Calls the Github Release Note API together with a release note configuration file (very similar to drafter but allows us to specify a previous tag)
3. Generates the notes for the current branch based on the last found tag.

This has the advantage that we can explicitly determine which release notes are generated from when to when and also use the contributors fields.

TODO:
1. Test on other repo
2. Make sure that the Merge commit in main that we used for testing disappears.


.github/workflows/release-drafter.yaml
```diff
@@ -31,12 +31,69 @@ jobs:
...
\ No newline at end of file
+          fetch-depth: 0
+          fetch-tags: true
+      - name: Setup Release with gh
+        env:
+          REF: ${{ github.ref }}
+          REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ steps.generate_token.outputs.token }}
+        run: |
+          git fetch --tags
```
hilmarf, line 51: Does the actions/checkout@v4 `fetch-tags: true` not work? Or why is there another `git fetch --tags` necessary?

.github/workflows/release-drafter.yaml
```diff
@@ -31,12 +31,69 @@ jobs:
...
+          notes=$(\
+             gh api  \
+               --method POST \
+               -H "Accept: application/vnd.github+json" \
+               -H "X-GitHub-Api-Version: 2022-11-28" \
+               /repos/${REPO}/releases/generate-notes \
+               -f "tag_name=${{env.RELEASE_VERSION}}" \
+               -f "target_commitish=${{ github.ref }}" \
+               -f "previous_tag_name=${previous_tag}" \
+               -f "configuration_file_path=.github/release.yml" \
```
hilmarf, line 77: When we do have the option to specify which config to use... can we then maybe place the config under `.github/config`?

hilmarf (dismissed):

.github/workflows/release-drafter.yaml
jakobmoellerdev: Leftover from testing, not necessary

.github/workflows/release-drafter.yaml
jakobmoellerdev: My original thought was that I would like to keep the default location as marked in the official documentation but im also fine to adopt it. WDYT?

hilmarf: @frewilhelm ?

jakobmoellerdev: adjusted. changed my mind :D

hilmarf (dismissed):

.github/workflows/release.yaml
```diff
@@ -181,8 +181,22 @@ jobs:
         git commit -m "Release $RELEASE_VERSION"
         msg="Release ${{ env.RELEASE_VERSION }}"
         git tag --annotate --message "${msg}" ${{ env.RELEASE_VERSION }}
-        # push both the tag as well as a release branch with that tag.
+        # push the tag.
         git push origin ${{ env.RELEASE_VERSION }}
+    # in case we do not have a release candidate, we should merge the release into the branch
+    # this is so that the tag shows up in the branch and we can use it for the release notes.
```
frewilhelm, line 187: This sound more complicated that is should be. What do we merge in which branch?

hilmarf (approved):

frewilhelm (approved):

PR #1140, opened 2024-11-26
feat: support splitting blobs when stored as OCI layer

mandelsoft:
There are OCI repositories with a layer size limitation. Because OCM potentially maps any external artifact to a single blob stored as layer on OCI registries, this could lead to problems.  An obvious problematic scenario is the transport of a multi-platform OCI image. Its blob format is an archive containing all images and all layers of those images.

This PR introduces the possibility to specify blob limits for OCI registries. The OCM-to-OCI mapping then
splits larger blobs into multiple layers.
The `localBlob` access method then uses a comma-separated list of blob  layer-blob digest to remember the sequence of layers. 
The access then combines the layerblobs again to a single stream.
 
Fixes: https://github.com/open-component-model/ocm-project/issues/12

A follow-up issue https://github.com/open-component-model/ocm-project/issues/338 describes the provisioning of appropriate defaults for common public registries.


api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigType, usage))
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigTypeV1))
+}
+
+// Config describes a memory based config interface
+// for configuring blob limits for underlying OCI manifest layers.
+type Config struct {
+	runtime.ObjectVersionedType `json:",inline"`
+	// BlobLimits describe the limit setting for host:port
+	// entries. As a spcial case (for testing) it is possible
```
fabianburth, line 26:
```suggestion
	// entries. As a special case (for testing) it is possible
```

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+	// to configure linits for CTF, also, by using "@"+filepath.
```
fabianburth, line 27:
```suggestion
	// to configure limits for CTF, also, by using "@"+filepath.
```

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+order:
+- highest priority first
+- longest matching sequence of component name segments first.
+
+If resolvers are defined, it is possible to use component version names on the
+command line without a repository. The names are resolved with the specified
+resolution rule.
+They are also used as default lookup repositories to lookup component references
+for recursive operations on component versions (<code>--lookup</code> option).
+`
```
fabianburth, line 128: copy paste mistake

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +124,107 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+		ref := strings.TrimSpace(c.refs[0])
+		_, c.data, err = c.m.namespace.GetBlobData(digest.Digest(ref))
+		if err != nil {
+			return 0, err
+		}
+		c.reader, err = c.data.Reader()
+		if err != nil {
+			return 0, err
+		}
+	}
```
Skarlso, line 216: This loop is weird. I'm not sure what the intention is. It never removes the `refs` so while there is an empty reader, this is potentially an endless loop.

Handling of EOF only happens if there was a read `if n > 0` you should check it before that in an error case.

You are using p to read but you have no idea if there is capacity for it. Why?

So:
- check eof in case of error
- remove refs
- reader close only happens when there was a read but not if it's eof or if there was an error.

api/ocm/cpi/repocpi/bridge_r.go
```diff
@@ -34,6 +34,16 @@ type RepositoryImpl interface {
 	io.Closer
 }
 
+type Chunked interface {
```
fabianburth, line 37: Although the purpose of this interface is quite obvious in the context of this PR, it might not be without it. Please add a brief comment.

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -183,32 +187,36 @@ func (c *ComponentVersionContainer) Update() (bool, error) {
...
 			if err != nil {
 				return false, fmt.Errorf("failed resource layer evaluation: %w", err)
 			}
-			if l > 0 {
-				layerAnnotations[l] = append(layerAnnotations[l], ArtifactInfo{
-					Kind:     ARTKIND_RESOURCE,
-					Identity: r.GetIdentity(desc.Resources),
-				})
-				layers.Delete(l)
+			if len(list) > 0 {
```
Skarlso, line 194: You don't need the length check if you are using a for loop.

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -183,32 +187,36 @@ func (c *ComponentVersionContainer) Update() (bool, error) {
...
 			if err != nil {
 				return false, fmt.Errorf("failed source layer evaluation: %w", err)
 			}
-			if l > 0 {
-				layerAnnotations[l] = append(layerAnnotations[l], ArtifactInfo{
-					Kind:     ARTKIND_SOURCE,
-					Identity: r.GetIdentity(desc.Sources),
-				})
-				layers.Delete(l)
+			if len(list) > 0 {
```
Skarlso, line 212: No need for the len check.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -88,7 +91,17 @@ func (m *localBlobAccessMethod) getBlob() (blobaccess.DataAccess, error) {
...
+
+	var (
+		data blobaccess.DataAccess
+		err  error
+	)
+	if len(refs) < 2 {
+		_, data, err = m.namespace.GetBlobData(digest.Digest(m.spec.LocalReference))
+	} else {
+		data = &composedBlock{m, refs}
+	}
```
fabianburth, line 104: why not handle the error locally within the if below line 101?

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +124,107 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+
+		_, data, err := c.m.namespace.GetBlobData(digest.Digest(ref))
+		if err != nil {
+			return nil, err
+		}
+		finalize.Close(data)
+		r, err := data.Reader()
+		if err != nil {
+			return nil, err
+		}
```
Skarlso, line 150: The finalizer logic here is misleading. This actually adds a finalizer that later will be called and not actually closing the reader. This logic should be apparent from the API. Call it `AddCloser` or something like that.

On top of that Close has a side-effect AND returns the finalizer which isn't being used here. Please try to stick with one paradigm in a method.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +124,107 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+}
+
+func (c *composedBlock) Reader() (io.ReadCloser, error) {
+	return &composedReader{
+		m:    c.m,
+		refs: c.refs,
+	}, nil
+}
+
+func (c composedBlock) Close() error {
```
Skarlso, line 171:
```suggestion
func (c *composedBlock) Close() error {
```

Just to be consistent. Even though it's `nil`.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +124,107 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+
+func (c *composedReader) Close() error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.reader != nil {
+		c.reader.Close()
+		c.data.Close()
+		c.reader = nil
+		c.refs = nil
```
Skarlso, line 227: Why is the `Close` cleaning the `refs`? 🤔

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -304,15 +325,49 @@ func (c *ComponentVersionContainer) AddBlob(blob cpi.BlobAccess, refName string,
 		return nil, errors.New("a resource has to be defined")
 	}
 
+	size := blob.Size()
+	var refs []string
+	if c.comp.repo.blobLimit > 0 && size != blobaccess.BLOB_UNKNOWN_SIZE && size > c.comp.repo.blobLimit {
+		reader, err := blob.Reader()
```
Skarlso, line 331: This reader is never closed.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
fabianburth: Besides, `p` is not large enough to store the entire data, the reader is closed afterwards and I cannot read the rest of it or am I interpreting the code wrong?

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -259,32 +267,45 @@ func (c *ComponentVersionContainer) Update() (bool, error) {
...
+		}
+		for _, ref := range refs {
+			d = &artdesc.Descriptor{Digest: digest.Digest(strings.TrimSpace(ref)), MediaType: media}
+			// find layer
+			layers := c.manifest.GetDescriptor().Layers
+			maxLen := len(layers) - 1
+			found := false
+			for i := range layers {
+				l := layers[len(layers)-1-i]
+				if i < maxLen && l.Digest == d.Digest && (d.Digest == "" || d.Digest == l.Digest) {
```
Skarlso, line 297: The second `&& (d.Digest == "" || d.Digest == l.Digest)` is redundant.

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -259,32 +267,45 @@ func (c *ComponentVersionContainer) Update() (bool, error) {
...
-				return s, len(layers) - 1 - i, nil
+		refs := strings.Split(a.LocalReference, ",")
+		media := a.GetMimeType()
+		if len(refs) > 1 {
+			media = mime.MIME_OCTET
+		}
+		for _, ref := range refs {
+			d = &artdesc.Descriptor{Digest: digest.Digest(strings.TrimSpace(ref)), MediaType: media}
+			// find layer
+			layers := c.manifest.GetDescriptor().Layers
```
Skarlso, line 292: The Descriptor here can be nil.

api/utils/accessio/chunkedreader.go
```diff
@@ -0,0 +1,118 @@
...
+// Once the reader reaches its limits it provides
+// a io.EOF.
+// It can be continued by Calling Next, which returns
+// whether a follow-up is required or not.
+type ChunkedReader struct {
+	lock   sync.Mutex
+	reader io.Reader
+	buffer *bytes.Buffer
+	size   int64
+	chunk  int
```
fabianburth, line 22: so here, chunk is the amount of chunks?

api/utils/accessio/chunkedreader.go
```diff
@@ -0,0 +1,118 @@
...
+	read   int64
+	err    error
+
+	preread uint
+}
+
+var _ io.Reader = (*ChunkedReader)(nil)
+
+func NewChunkedReader(r io.Reader, chunk int64, preread ...uint) *ChunkedReader {
```
fabianburth, line 31: and here, chunk is the size of the chunks?

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -304,15 +325,49 @@ func (c *ComponentVersionContainer) AddBlob(blob cpi.BlobAccess, refName string,
 		return nil, errors.New("a resource has to be defined")
 	}
 
+	size := blob.Size()
+	var refs []string
+	if c.comp.repo.blobLimit > 0 && size != blobaccess.BLOB_UNKNOWN_SIZE && size > c.comp.repo.blobLimit {
```
Skarlso, line 330: Hm, since you have warding code in chunk for concurrency, I'm guessing this can be called concurrently. Meaning `blobLimit` here, could potentially be changed mid execution.

I suggest putting it outside as a variable like `size` and use that instead of keep calling it.

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+func init() {
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigType, usage))
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigTypeV1))
+}
+
+// Config describes a memory based config interface
+// for configuring blob limits for underlying OCI manifest layers.
+type Config struct {
```
fabianburth, line 23: We should be able to provide sane defaults for some popular public registries.

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+	runtime.ObjectVersionedType `json:",inline"`
+	// BlobLimits describe the limit setting for host:port
+	// entries. As a spcial case (for testing) it is possible
+	// to configure linits for CTF, also, by using "@"+filepath.
+	BlobLimits BlobLimits `json:"blobLimits"`
+}
+
+type BlobLimits map[string]int64
```
Skarlso, line 31: I wonder. I've seen, and please correct me if I'm wrong, that this could potentially be called concurrently?

In that case, this will race.

api/utils/blobaccess/chunked/access.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		t.Close()
+		return nil, err
+	}
+	return t.AsBlob(mime.MIME_OCTET), nil
+}
+
+// ChunkedBlobSource provides a sequence of
+// bpi.BlobAccess objects.
+type ChunkedBlobSource interface {
```
fabianburth, line 33: Couldn't this be implemented with `io.LimitedReader` instead of writing our own `ChunkedReader`?

api/utils/accessio/chunkedreader.go
```diff
@@ -0,0 +1,118 @@
...
+
+	if c.read >= c.size {
+		return 0, io.EOF
+	}
+	if c.read+int64(len(p)) > c.size {
+		p = p[:c.size-c.read] // read at most rest of chunk size
+	}
+	if c.buffer != nil && c.buffer.Len() > 0 {
+		// first, consume from buffer
+		n, _ := c.buffer.Read(p)
```
Skarlso, line 51: Don't ignore the read error please.

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: appropriate doc provided

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: you are right, refs is never shrinked, strange, why are the tests succeeding

mandelsoft: ahh, another method, streaming is not tested :-(

mandelsoft: The final reader close is potentially done with the external `Close` of the `composedReader`

mandelsoft: no data get, streamed read and steamed read with small buffer is tested.

api/ocm/cpi/repocpi/bridge_r.go
mandelsoft: done

api/ocm/extensions/repositories/genericocireg/componentversion.go
mandelsoft: done

api/ocm/extensions/repositories/genericocireg/componentversion.go
mandelsoft: done

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: Because the else was different before :-),  fixed...

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -259,32 +267,45 @@ func (c *ComponentVersionContainer) Update() (bool, error) {
...
+			media = mime.MIME_OCTET
+		}
+		for _, ref := range refs {
+			d = &artdesc.Descriptor{Digest: digest.Digest(strings.TrimSpace(ref)), MediaType: media}
+			// find layer
+			layers := c.manifest.GetDescriptor().Layers
+			maxLen := len(layers) - 1
+			found := false
+			for i := range layers {
+				l := layers[len(layers)-1-i]
```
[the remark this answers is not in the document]
mandelsoft: the for loop would be more complicated. I don't know what is better

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: The return value is intended to be chainable.

The Finalizer is not touched, here and therefore the interface should not be changed in the PR.
Anyway: `finalize.Close(x)` directly say what is done. The finalizer will close the given object.
Maybe, we could name the method `CloseOnFinalize`. Or, to be consistent with other other such methods `WithCloser`.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: yes, that's always the IDE.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: To be able to return ErrClosed, (what is unfortunately missing)

changed.

api/ocm/extensions/repositories/genericocireg/componentversion.go
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/componentversion.go
mandelsoft: yes

api/utils/accessio/chunkedreader.go
mandelsoft: yes, changed

api/utils/accessio/chunkedreader.go
mandelsoft: yes, changed

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
Skarlso: Hum, either is okay. But yeah, don't have to do that here I guess. :)

api/ocm/extensions/repositories/genericocireg/componentversion.go
mandelsoft: you are right, fixed.

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: I added an appropriate defaulting with a default registry. The value given, here, can be overwritten
by the other specification methods.

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,128 @@
...
+
+	l, ok := b[hostport]
+	if ok {
+		return l
+	}
+	l, ok = b[host]
+	if ok {
+		return l
+	}
+	return -1
```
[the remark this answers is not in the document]
mandelsoft: you are completely right, but there are many other places not touched by this PR. Therefore, I propose to use a separate PR for this, covering all places. 

But I fixed it, here, together with your improvement proposal.

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: yes, but the config object should be configured before adding it to the config context.
Modifying config object once handled by the config context will always lead to unpredictable results
and is there not allowed.

api/utils/blobaccess/chunked/access.go
mandelsoft: yes and no, the `LimitedReader` stops after N bytes, but cannot be continued.
The `ChunkedReader` offers a `Next` method to continue reading the next chunk after an EOF.
Additionally, it provides a check, whether the actual chunk is already done.

api/utils/accessio/chunkedreader.go
mandelsoft: Hmm, the buffer `Read`only returns an EOF error, which is handled later.
But I'll replace the length check by an error check.

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
Skarlso: nice, thanks

api/ocm/extensions/repositories/genericocireg/config/type.go
Skarlso: Okay cool. Thanks for the explanation!

api/ocm/extensions/repositories/genericocireg/config/type.go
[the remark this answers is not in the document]
Skarlso: Sounds good to me 👍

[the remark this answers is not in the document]
Skarlso: https://github.com/open-component-model/ocm/issues/1153

api/ocm/extensions/repositories/genericocireg/componentversion.go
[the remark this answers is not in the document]
Skarlso: I think this looks a lot nicer. :) Well done.

api/utils/blobaccess/chunked/access.go
fabianburth: but since we wrap the chunk readers next in the blob accesses next, couldn't we set up the consecutive limited reader in the blob accesses next?

mandelsoft: Unfortunately the Limited Reader does not provide information about the final EOF. If the number of bytes read, when EOF is provided, gives this hint, that the steam has been finished. But otherwise we always have to cover the case that the size is a multiple of the chunk size.  Therefore, we have to look ahead in the original stream.

This has been separated in a LookAheadReader now, which significantly reduces the complexity in the ChunkedReader.

api/utils/accessio/lookaheadreader.go
```diff
@@ -0,0 +1,74 @@
...
+		return n, nil
+	}
+
+	cnt, err := r.reader.Read(p[n:])
+	return cnt + n, err
+}
+
+// LookAhead  provides a preview of upcoming data.
+// It tries to fill the complete given buffer.
+// The regular data stream provided by ead is not affected.
```
fabianburth, line 52:
```suggestion
// The regular data stream provided by read is not affected.
```

typo

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +125,119 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+
+type composedReader struct {
+	lock   sync.Mutex
+	m      *localBlobAccessMethod
+	refs   []string
+	reader io.ReadCloser
+	data   blobaccess.DataAccess
+}
+
+func (c *composedReader) Read(p []byte) (n int, err error) {
```
fabianburth, line 184: what's the purpose of the named return values here?

mandelsoft: ??? That's the standard reader interface.

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -304,15 +321,51 @@ func (c *ComponentVersionContainer) AddBlob(blob cpi.BlobAccess, refName string,
...
+			if b == nil {
+				break
+			}
+			err = c.addLayer(b, &refs)
+			b.Close()
+
+			if err != nil {
+				return nil, errors.Wrapf(err, "chunked blob")
+			}
+		}
```
fabianburth, line 348: couldnt we replace that with something like this
```
		for {
			br, err := func() (bool, error) {
				f, err := blobaccess.NewTempFile("", "chunk-*", vfsattr.Get(c.GetContext()))
				if err != nil {
					return true, err
				}
				defer f.Close()
				written, err := io.CopyN(f.Writer(), reader, limit)
				if err != nil && err != io.EOF {
					if err := f.Close(); err != nil {
						errors.Join(err, errors.Wrapf(err, "close temp file"))
					}
					return true, err
				}
				if written != 0 {
					b := f.AsBlob(blob.MimeType())
					err = c.addLayer(b, &refs)
					if err != nil {
						return true, err
					}
				}
				if err == io.EOF {
					return true, nil
				}
				return false, nil
			}()
			if err != nil {
				return nil, err
			}
			if br {
				break
			}
		}
```
and get rid of the `ChunkedReader`, the `LookAheadReader` and the `ChunkedBlobAccess`?

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: I'll add the var block.

api/ocm/extensions/repositories/genericocireg/componentversion.go
fabianburth: sure, could also do something like this

```
BlobAccessForChunk := func(r io.Reader, chunk int64) (cpi.BlobAccess, bool, error) {
	f, err := blobaccess.NewTempFile("", "chunk-*", vfsattr.Get(c.GetContext()))
	if err != nil {
		return nil, true, err
	}
	defer f.Close()
	written, err := io.CopyN(f.Writer(), reader, limit)
	if err != nil && err != io.EOF {
		return nil, true, err
	}
	if written <= 0 {
		return nil, true, fmt.Errorf("no data written")
	}
	return f.AsBlob(blob.MimeType()), err == io.EOF, nil
}

for {
	b, br, err := BlobAccessForChunk(reader, limit)
	if err != nil {
		return nil, err
	}
	defer b.Close()

	err = c.addLayer(b, &refs)
	if err != nil {
		return nil, err
	}
	if br {
		break
	}
}
```

api/ocm/cpi/repocpi/bridge_r.go
```diff
@@ -34,6 +34,24 @@ type RepositoryImpl interface {
...
+// Chunked is an optional interface, which
+// may be implemented to accept a blob limit for mapping
+// local blobs to an external storage system.
+type Chunked interface {
+	// SetBlobLimit sets the blob limit if possible.
+	// It returns true, if this was successful.
+	SetBlobLimit(s int64) bool
+}
+
+// SetBlobLimit tries to set a blob limt for a repository
```
fabianburth, line 46:
```suggestion
// SetBlobLimit tries to set a blob limit for a repository
```
typo

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +125,119 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		err = finalize.Finalize()
+		if err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), nil
+}
```
fabianburth, line 163: Would it make sense to replace this with an implementation using io.MultiReader()?

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -111,3 +125,119 @@ func (m *localBlobAccessMethod) Get() ([]byte, error) {
...
+	return nil
+}
+
+type composedReader struct {
+	lock   sync.Mutex
+	m      *localBlobAccessMethod
+	refs   []string
+	reader io.ReadCloser
+	data   blobaccess.DataAccess
+}
```
fabianburth, line 182: Same question here, would it make sense to replace this with an implementation using io.MultiReader()?

api/ocm/extensions/repositories/genericocireg/bloblimits.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// Add limits for known OCI repositories, here,
+	// or provide init functions in specialized packages
+	// by calling AddDefaultBlobLimit.
+}
+
+// AddDefaultBlobLimit can be used to set default blob limits
+// for known repositories.
+// Those limits will be overwritten, by blob limits
+// given by a configuration ovject and the repository
```
fabianburth, line 26:
```suggestion
// given by a configuration object and the repository
```

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,109 @@
...
+
+If blob limits apply to a registry, local blobs with a size larger than
+the configured limit will be split into several layers with a maximum
+size of the given value.
+
+These settings can be overwritten by explicit settings in an OCM
+repository specification for those repositories.
+
+The most specific entry will be used. If a registry with a dedicated
+port is requested, but no explicit such configuration is found, the
```
fabianburth, line 107:
```suggestion
port is requested, but no such explicit configuration is found, the
```

api/ocm/extensions/repositories/genericocireg/repository.go
```diff
@@ -44,23 +48,70 @@ type RepositoryImpl struct {
 	nonref   cpi.Repository
 	ocirepo  oci.Repository
 	readonly bool
+	// blobLimit is the size limit for layers maintained for the storage of localBlobs.
+	// The value -1 means an unconfigured value,
+	// a value == 0 disables the limiting and
+	// a value > 0 enabled the usage of the specified size.
```
fabianburth, line 54: So:
-1: if a blob limit configuration is found for the host:port, it's used
== 0: even if a blob limit configuration is found for the host:port, it's ignored
> 0: if a blob limit configuration is found for the host:port, it won't be applied but the concrete value assigned here will be used

that right?

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: Basically yes, but a MultiReader requires a slice of preopened readers, and here the readers (or better blob accesses ) are created one after the other.

fabianburth: Sure, but are we expecting that opening the readers in parallel would cause issues here?

api/ocm/extensions/repositories/genericocireg/bloblimits.go
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/config/type.go
```diff
@@ -0,0 +1,109 @@
...
+
+const usage = `
+The config type <code>` + ConfigType + `</code> can be used to set some
+blob layer limits for particular OCI registries used to host OCM repositories;
+
+<pre>
+    type: ` + ConfigType + `
+    blobLimits:
+        dummy.io: 65564
+        dummy.io:8443: 32768
```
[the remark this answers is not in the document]
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/config/type.go
mandelsoft: fixed

api/ocm/extensions/repositories/genericocireg/repository.go
mandelsoft: clarified

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
mandelsoft: completely reworked

api/ocm/extensions/repositories/genericocireg/repository.go
```diff
@@ -44,23 +48,70 @@ type RepositoryImpl struct {
 	nonref   cpi.Repository
 	ocirepo  oci.Repository
 	readonly bool
+	// blobLimit is the size limit for layers maintained for the storage of localBlobs.
+	// The value -1 means an unconfigured value (a default from the blob limit configuration is used),
+	// a value == 0 disables the limiting and (a default from the blob limit configuration is ignored),
+	// a value > 0 enabled the usage of the specified size.
+	blobLimit int64
```
fabianburth, line 55: Perhaps you should also use this description in line 66-69.

api/ocm/extensions/repositories/genericocireg/bloblimits.go
```diff
@@ -0,0 +1,46 @@
...
+	lock              sync.Mutex
+)
+
+func init() {
+	defaultBlobLimits = config.BlobLimits{}
+
+	// Add limits for known OCI repositories, here,
+	// or provide init functions in specialized packages
+	// by calling AddDefaultBlobLimit.
+}
```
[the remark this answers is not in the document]
mandelsoft: set for ghcr

api/ocm/extensions/repositories/genericocireg/repository.go
mandelsoft: done

docs/reference/ocm_configfile.md
```diff
@@ -24,6 +24,27 @@ The following configuration types are supported:
...
   </pre>
+- <code>blobLimits.ocireg.ocm.config.ocm.software</code>
+  The config type <code>blobLimits.ocireg.ocm.config.ocm.software</code> can be used to set some
+  blob layer limits for particular OCI registries used to host OCM repositories;
+
+  <pre>
+      type: blobLimits.ocireg.ocm.config.ocm.software
+      blobLimits:
+          dummy.io: 65564
+          dummy.io:8443: 32768
```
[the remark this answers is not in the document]
mandelsoft: done

fabianburth (approved):

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -299,20 +317,74 @@ func (c *ComponentVersionContainer) GetStorageContext() cpi.StorageContext {
 	return ocihdlr.New(c.comp.GetName(), c.Repository(), c.comp.repo.ocirepo.GetSpecification().GetKind(), c.comp.repo.ocirepo, c.comp.namespace, c.manifest)
 }
 
+func blobAccessForChunk(blob blobaccess.BlobAccess, fs vfs.FileSystem, r io.Reader, limit int64) (cpi.BlobAccess, bool, error) {
+	f, err := blobaccess.NewTempFile("", "chunk-*", fs)
```
Skarlso, line 321: This `f` is never closed in case of a success.

Skarlso (approved):
LGTM with a single comment.

api/ocm/extensions/repositories/genericocireg/componentversion.go
fabianburth: I think that's because the file does not have a view. But the blob (`f.AsBlob(...)`) that is returned is based on that file and also closes that file. So you hand the close responsibility over to the blob receiver.

PR #1141, opened 2024-11-26
docs: release process - add follow up section

hilmarf:

Once the release has been created, there are some follow up actions to keep in mind.


RELEASE_PROCESS.md
```diff
@@ -244,3 +244,27 @@ branches were created on demand.
 This model has now largely been replaced by the model we see in
 this document. However you might still encounter leftovers form the old model
 and you are encouraged to create issues regarding inconsistencies.
+
+## Follow up actions
```
jakobmoellerdev, line 248: Should we link to the corresponding step that needs to succeed here or even implicitly put that as a comment on the action run?

RELEASE_PROCESS.md
```diff
@@ -244,3 +244,29 @@ branches were created on demand.
...
+
+Review and merge open pull requests.
```
frewilhelm, line 250: nit: This sounds like another subtitle (is this necessary)

hilmarf: any suggestion?

```suggestion
Review and merge open pull requests at the following locations.
```

jakobmoellerdev (approved):

PR #1145, opened 2024-11-26
chore: autocollapse release notes

jakobmoellerdev:

Adds a small script that autocollapses our release notes again because we no longer use release drafter


hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+
+# Auto Collapse
+#This is a script that takes in
+# - a markdown file
+# - a designated output file
+# - the number of lines as a threshold on when to collapse a section.
+#
+# Sample
+# ./auto_collapse.sh test_example.md test_example_collapsible.md 10
```
fabianburth, line 10:
```suggestion
# ./auto_collapse.sh test_example.md test_example_collapsible.md 5
```
That's absolute nitpicking, but just to avoid confusion, since the example file does not have a list that's 10 entries long and therefore, the script has no effect.

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
fabianburth, line 1: file might need a `chmod +x`

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+# Sample
+# ./auto_collapse.sh test_example.md test_example_collapsible.md 10
+
+# Input and output files
+INPUT_FILE=${1:-"README.md"}
+OUTPUT_FILE=${2:-"README_collapsible.md"}
+THRESHOLD=${3:-10}
+
+# Ensure output file is empty initially
+rm -f "$OUTPUT_FILE"
```
frewilhelm, line 18: Instead of `force` removing a file, you could also just write to the file using `>` to overwrite previous content.
(I am really not a fan of `-f` in scripts on paths that are passed through variables/input parameters)

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Variables to track sections
+inside_section=false
+section_lines=()
+section_header=""
+
+INPUT="$(cat "$INPUT_FILE")"
+FULL_CHANGELOG=$(grep -E "Full Changelog" < "$INPUT_FILE")
```
frewilhelm, line 26: Why do you need extended regex in this case?

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+
+if [[ -z $FULL_CHANGELOG ]]; then
+    echo "Full Changelog not found in the input file."
+else
+    echo "Full Changelog found in the input file."
+    INPUT=$(echo "$INPUT" | sed '/Full Changelog/d')
+fi
+
+# Function to count changes (lines starting with '*')
+count_changes() {
```
frewilhelm, line 36: `grep "^*" $INPUT_FILE | wc -l` does the same job and looks a bit easier (and we avoid array expansion in bash :D). What do you think?

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+}
+
+# Function to process and write a section
+write_section() {
+    local header="$1"
+    local lines=("${@:2}")
+    num_changes=$(count_changes "${lines[@]}")
+
+    # Write the section header as is
+    echo "$header" >> "$OUTPUT_FILE"
```
frewilhelm, line 54: Here we could use the aforementioned `>` to overwrite the previous file

hack/collapse/auto_collapse.sh
frewilhelm: If it is on a variable and not a file, you can use `grep "^*" <<< "${lines}" | wc -l`

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+section_header=""
+
+INPUT="$(cat "$INPUT_FILE")"
+FULL_CHANGELOG=$(grep -E "Full Changelog" < "$INPUT_FILE")
+
+if [[ -z $FULL_CHANGELOG ]]; then
+    echo "Full Changelog not found in the input file."
+else
+    echo "Full Changelog found in the input file."
+    INPUT=$(echo "$INPUT" | sed '/Full Changelog/d')
```
frewilhelm, line 32:
```suggestion
    INPUT=$(sed '/Full Changelog/d' <<< "${INPUT}")
```

hack/collapse/auto_collapse.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Ensure output file is empty initially
+rm -f "$OUTPUT_FILE"
+
+# Variables to track sections
+inside_section=false
+section_lines=()
+section_header=""
+
+INPUT="$(cat "$INPUT_FILE")"
```
frewilhelm, line 25: Is there a reason why this variable only consists of capital letters?

hack/collapse/auto_collapse.sh
jakobmoellerdev: you should be able to exec as is

hack/collapse/test_auto_collapse.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
frewilhelm, line 3: This looks very complicated. What do you want to achieve?

hack/collapse/test_auto_collapse.sh
```diff
@@ -0,0 +1,26 @@
...
+export INPUT_FILE="$SCRIPT_DIR/test_example.md"
+export OUTPUT_FILE="$SCRIPT_DIR/test_example_collapsible.md"
+export EXPECTED_FILE="$SCRIPT_DIR/test_example_expected_collapsible.md"
+export SCRIPT="$SCRIPT_DIR/auto_collapse.sh"
+
+# Run the script
+bash "$SCRIPT" "$INPUT_FILE" "$OUTPUT_FILE" 5
+
+if [ ! -f "$OUTPUT_FILE" ]; then
+    echo "Test failed: Output file not found."
```
frewilhelm, line 15: nitpicking but interesting: When a string is passed in `"`, bash will look into the string if there is anything to expand. Using `'` will tell bash to not look into the string because we do not want anything to expand. Thus, using `'` for plain strings is a bit more performant (not relevant for this script)

hack/collapse/auto_collapse.sh
frewilhelm: Can you try this `num_changes="$(IFS=$'\n'; grep "^*" <<< "${lines[*]}" | wc -l)"`

PR #1146, opened 2024-11-27
feat: support for feature gates

mandelsoft:
This PR introduces the concept of feature gates for the CM library/CLI.

A feature has a name and a setting.
The standard setting is a `mode`. Setting the mode to `off` enforces the feature to be disabled.
All other mode values enable the feature and can be used by the implementation to select the way it is supported.
If a feature is not explicitly configured to have a setting it is assumed to be disabled by default.

Optionally, settings may include a arbitrary set of  name/value pairs, a value might be structured.

A feature may be enabled or disabled by default, depending of the arguments to the 
feature query functions.

The set of feature gates is a context attribute `ocm.software/ocm/feature-gates` with the short name
`featuregates`.

Additionally, a dedicated config object is provided ( `featuregates.config.ocm.software`).
It supports the following config fields:

- `features`: *map[string]`FeatureGate`*  map of feature settings

A `FeatureGate` has the fields:
- `mode`: *string* if set to `off` features is enforced to be disabled. Other values enable the feature
- `attributes`: *map[string]<any json/yaml>*  optional named feature settings


api/datacontext/attrs/featuregatesattr/attr.go
```diff
@@ -0,0 +1,199 @@
...
+
+	Features map[string]*FeatureGate `json:"features"`
+}
+
+// FeatureGate store settings for a particular feature gate.
+// To be extended by additional config possibility.
+// Default behavior is to be enabled if entry is given
+// for a feature name and mode is not equal *off*.
+type FeatureGate struct {
+	Mode       string                     `json:"mode"`
```
jakobmoellerdev, line 83: Generally in favor of this, but why choose to have a Mode setting for the Gate if you already have attributes? Shouldn't the mere presence of the gate determine if its enabled or disabled?

mandelsoft: The idea of the mode is to have a fixed/non-optional config field usable to enable/disable a feature and control an implementation mode (if required).

The attributes are completely optional and intended to pass arbitrary information to the feature. (An on/off switch does not use attributes at all (omitempty))

jakobmoellerdev: > The idea of the mode is to have a fixed/non-optional config field usable to enable/disable a feature and control an implementation mode (if required). The value off is always used to enforce disabling the feature (in case it is enabled by default).

Could you give me an example for an "implementation mode" that is not on/off? Ive never seen it anywhere in production and was assuming that even if there was a non on/off, it would be part of the attributes.

api/datacontext/config/featuregates/type.go
```diff
@@ -0,0 +1,69 @@
...
+
+func init() {
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigType, usage))
+	cfgcpi.RegisterConfigType(cfgcpi.NewConfigType[*Config](ConfigTypeV1, usage))
+}
+
+type FeatureGate = featuregatesattr.FeatureGate
+
+// Config describes a memory based repository interface.
+type Config struct {
```
jakobmoellerdev, line 22: Whats the reason to have a config and an attribute set, shouldnt attributes alone be enough?

cmds/ocm/commands/ocmcmds/common/handlers/featurehdlr/sort.go
```diff
@@ -0,0 +1,17 @@
...
+
+func Compare(a, b interface{}) int {
+	aa := a.(*Object)
+	ab := b.(*Object)
+
+	return strings.Compare(aa.Name, ab.Name)
+}
+
+// Sort is a processing chain sorting original objects provided by type handler.
+var Sort = processing.Sort(Compare)
```
hilmarf, line 17: out of the blue...

cmds/ocm/commands/ocmcmds/common/handlers/featurehdlr/suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestConfig(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Version handler Test Suite")
```
hilmarf, line 12:
```suggestion
	RunSpecs(t, "feature handler Test Suite")
```

api/datacontext/attrs/featuregatesattr/attr.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/mandelsoft/goutils/general"
+	"sigs.k8s.io/yaml"
+
+	"ocm.software/ocm/api/datacontext"
+	"ocm.software/ocm/api/utils/runtime"
+)
+
+const (
+	ATTR_KEY   = "ocm.software/ocm/feature-gates"
+	ATTR_SHORT = "featuregates"
```
hilmarf, line 17:
```suggestion
	ATTR_SHORT = "featuregates"
	ATTR_KEY   = "ocm.software/ocm/" + ATTR_SHORT
```

docs/reference/ocm_ocm.md
```diff
@@ -42,4 +43,5 @@ ocm ocm [<options>] <sub command> ...
 * [ocm <b>ocm-accessmethods</b>](ocm_ocm-accessmethods.md)	 &mdash; List of all supported access methods
 * [ocm <b>ocm-downloadhandlers</b>](ocm_ocm-downloadhandlers.md)	 &mdash; List of all available download handlers
 * [ocm <b>ocm-references</b>](ocm_ocm-references.md)	 &mdash; notation for OCM references
+* [ocm <b>ocm-references</b>](ocm_ocm-references.md)	 &mdash; notation for OCM references
```
hilmarf, line 46: This looks like a bug! Why is 'ocm-references' listed now twice?

cmds/ocm/commands/ocmcmds/featuregates/cmd.go
```diff
@@ -0,0 +1,26 @@
...
+	"ocm.software/ocm/cmds/ocm/commands/ocmcmds/names"
+	"ocm.software/ocm/cmds/ocm/common/utils"
+)
+
+var Names = names.FeatureGates
+
+// NewCommand creates a new command.
+func NewCommand(ctx clictx.Context) *cobra.Command {
+	cmd := utils.MassageCommand(&cobra.Command{
+		Short: "Commands acting on actions",
```
hilmarf, line 18:
```suggestion
		Short: "Commands acting on feature gates",
```

docs/reference/ocm_get_featuregates.md
```diff
@@ -0,0 +1,50 @@
...
+Show feature gates and the activation.
+
+The following feature gates are supported:
+
+
+With the option <code>--output</code> the output mode can be selected.
+The following modes are supported:
+  - <code></code> (default)
+  - <code>JSON</code>
+  - <code>json</code>
```
hilmarf, line 34: what's the difference here?

cmds/ocm/commands/ocmcmds/common/handlers/featurehdlr/typehandler_test.go
hilmarf, line 1: nonsense - just copy pasted from vershdlr

hilmarf (approved):

PR #1160, opened 2024-11-28
chore: make sure that dockerfiles are setup in a standard way

jakobmoellerdev:

This simplifies the docker images update process because it runs through dependabot.yml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-ARG GO_VERSION="1.23"
-ARG ALPINE_VERSION="3.20"
-ARG DISTROLESS_VERSION=debian12:nonroot@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
-
-FROM golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS build
+FROM golang:1.23-alpine3.20 AS build
 
 WORKDIR /src
 RUN go env -w GOMODCACHE=/root/.cache/go-build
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,21 +17,21 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     "-s -w -X ocm.software/ocm/api/version.gitVersion=$VERSION -X ocm.software/ocm/api/version.buildDate=$NOW" \
     -o /bin/ocm ./cmds/ocm/main.go
 
-FROM gcr.io/distroless/static-${DISTROLESS_VERSION}
+FROM gcr.io/distroless/static-debian12:nonroot@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
 # pass arg from initial build
 ARG DISTROLESS_VERSION
```
frewilhelm, line 22: This is not required anymore

components/ocmcli/Dockerfile
```diff
@@ -1,7 +1,6 @@
 # This Dockerfile is used by `make` via the Makefile
 
-ARG DISTROLESS_VERSION=debian12:nonroot@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
-FROM --platform=$BUILDPLATFORM gcr.io/distroless/static-${DISTROLESS_VERSION}
+FROM --platform=$BUILDPLATFORM gcr.io/distroless/static-debian12:nonroot@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
 # pass arg from initial build
 ARG DISTROLESS_VERSION
```
frewilhelm, line 5:
```suggestion
```

Dockerfile
frewilhelm, line 22:
```suggestion
```

frewilhelm (changes requested):

frewilhelm (approved):

PR #1163, opened 2024-11-28
chore: rework default Dockerfile

jakobmoellerdev:

uses make now to unify our build process to work exactly like we expect, everytime


Dockerfile
```diff
@@ -1,25 +1,24 @@
...
-    -o /bin/ocm ./cmds/ocm/main.go
+
+ENV BUILD_FLAGS="-trimpath"
+
+RUN make bin/ocm
 
 FROM gcr.io/distroless/static-debian12:nonroot@sha256:d71f4b239be2d412017b798a0a401c44c3049a3ca454838473a4c32ed076bfea
 
-COPY --from=build /bin/ocm /usr/local/bin/ocm
+COPY --from=build /src/bin/ocm /usr/local/bin/ocm
```
frewilhelm, line 21: Is an absolute path necessary? What about `bin/ocm`?

Makefile
```diff
@@ -26,7 +26,10 @@ SOURCES := $(shell go list -f '{{$$I:=.Dir}}{{range .GoFiles }}{{$$I}}/{{.}} {{e
 GOPATH                                         := $(shell go env GOPATH)
 
 NOW         := $(shell date -u +%FT%T%z)
-BUILD_FLAGS := "-s -w \
+ifneq ($(BUILD_FLAGS),)
+BUILD_FLAGS := $(BUILD_FLAGS)
```
frewilhelm, line 30: This looks odd^^. If `BUILD_FLAGS` is set why do we set it again?

Dockerfile
jakobmoellerdev: /src is the working directory from the other image, it is not given by default and the makefile by default puts it under /bin/ocm but we change the WORKDIR to /src

frewilhelm: I don't understand this. You say that the makefile puts the binary in `/bin/ocm`. Thats weird. I would have expected it in `bin/ocm`. Why is changing the `WORKDIR` to `/src` causing the binary to be in `/src/bin/ocm` Does the makefile consider the working-directory?

PR #1165, opened 2024-11-28
chore: rework default Dockerfile

jakobmoellerdev:

uses make now to unify our build process to work exactly like we expect, everytime

follows up on https://github.com/open-component-model/ocm/pull/1163 as github went crazy due to me reverting to a previously detached commit


components/ocmcli/Makefile
```diff
@@ -68,40 +75,35 @@ $(GEN)/build: $(GEN) $(GEN)/.exists $(CMDSRCS) $(OCMSRCS)
...
-        echo building platform $$i; \
-	docker buildx build --load -t $(IMAGE):$(VERSION)-$$tag --platform $$i --file Dockerfile $(REPO_ROOT) \
-          --build-arg OCM_VERSION=$(EFFECTIVE_VERSION) \
-          --build-arg SUFFIX=$$tag; \
+		tag=$$(echo $$i | sed -e s:/:-:g); \
+			echo building platform $$i; \
+		docker buildx build --load -t $(IMAGE):$(VERSION)-$$tag --platform $$i --file $(REPO_ROOT)/Dockerfile $(REPO_ROOT) \
+			--build-arg OCM_VERSION=$(EFFECTIVE_VERSION) \
+			--label org.opencontainers.image.version=$(VERSION) \
+			--label org.opencontainers.image.revision=$(VERSION) $(ADDITIONAL_BUILDX_ARGS); \
```
frewilhelm, line 100:
```suggestion
			--label org.opencontainers.image.revision=$(VERSION) $(ADDITIONAL_BUILDX_ARGS)
```

These should not be required, right?

jakobmoellerdev: you need this to parallelize the build instruction

frewilhelm: interesting. Thanks for clarification

frewilhelm (approved):

Issue #1174, opened 2024-12-06
Replace module https://pkg.go.dev/github.com/juju/fslock with something else because of LGPL license


morri-son:
**What would you like to be added**:
https://github.com/open-component-model/ocm/blob/main/api/utils/filelock/lock.go uses package https://pkg.go.dev/github.com/juju/fslock. This package is licensed under LPGL which is very restrictive. We must replace this usage with  different library, e.g. 
https://github.com/rogpeppe/go-internal/blob/master/lockedfile/lockedfile.go

**Why is this needed**:
Static linking with LGPL libraries changes the licensing landscape significantly:

**LGPL Static Linking Requirement:**

When statically linking LGPL libraries, you must either:

- Provide the full source code of your application
- Provide a mechanism for users to replace the library
- Potentially release your entire application under LGPL

**Practical Implications:**

- This could require open-sourcing your entire application
- Significant compliance burden for commercial use
- Potential legal risks if not handled correctly

**Tasks

- [ ] Replace library
- [ ] Update ./NOTICE file in the repository and remove the part for the LGPL lib

PR #1176, opened 2024-12-06
fix: replace juju with rogpeppe go internal

Skarlso:

Fixes https://github.com/open-component-model/ocm/issues/1174


Skarlso:
This might be a problem for you @jakobmoellerdev in https://github.com/open-component-model/ocm/pull/869.

hilmarf (approved):

PR #1184, opened 2024-12-10
feat: replace docker client code with regclient

Skarlso:


api/tech/oras/client.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (c *Client) Fetch(ctx context.Context, desc ociv1.Descriptor) (io.ReadCloser, error) {
+	src, err := c.createRepository(c.Ref)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve ref %q: %w", c.Ref, err)
+	}
+
+	// oras requires a Resolve to happen before a fetch because
+	// -1 is an invalid size.
```
jakobmoellerdev, line 116: maybe explain how -1 can happen in the first place here, its not apparent to me

api/tech/oras/client.go
```diff
@@ -0,0 +1,173 @@
...
+		// that layer resulting in the created tag pointing to the right
+		// blob data.
+		if err := repository.PushReference(ctx, d, reader, c.Ref); err != nil {
+			return fmt.Errorf("failed to push tag: %w", err)
+		}
+
+		return nil
+	}
+
+	// We have a digest, so we push use plain push for the digest.
```
jakobmoellerdev, line 100:
```suggestion
	// We have a digest, so we use plain push for the digest.
```
nit

api/tech/oras/client.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+
+	repository, err := c.createRepository(c.Ref)
+	if err != nil {
+		return err
+	}
+
+	if split := strings.Split(c.Ref, ":"); len(split) == 2 {
```
jakobmoellerdev, line 88: shouldnt the reference have an IsTag() method somewhere? maybe this is too low level already here, so feel free to ignore if it doesnt apply.

jakobmoellerdev (commented):
Mostly nits, generally LGTM but needs some clarification regarding concurrent access of readers

api/tech/oras/client.go
Skarlso: You're right. I'll link in the issue.

api/tech/oras/client.go
Skarlso: It should but it doesn't. 🤣🤣

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -126,57 +126,54 @@ func (r *RepositoryImpl) getResolver(comp string) (resolve.Resolver, error) {
...
-					}(),
+				if rootCAs == nil {
+					rootCAs = rootcertsattr.Get(r.GetContext()).GetRootCertPool(true)
 				}
+				return rootCAs
 			}(),
-		})),
+		}
+		client = &http.Client{
+			Transport: retry.NewTransport(&http.Transport{
```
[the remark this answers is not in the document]
Skarlso: Oh interesting. Thanks for that.

api/tech/oras/client.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("failed to list tags: %w", err)
+	}
+
+	return result, nil
+}
+
+func (c *Client) createRepository(ref string) (*remote.Repository, error) {
+	src, err := remote.NewRepository(ref)
```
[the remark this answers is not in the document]
Skarlso: I was thinking about actually putting THIS thing into the struct instead of constructing it.

But luckily it's cheap as dirt, so it's fine to be called over and over again. And I have a greater confidence in storing the ref instead of the repository. Also, storying the Repository pointer is more costly in memory.

api/tech/oras/client.go
Skarlso: https://github.com/oras-project/oras-go/issues/822

basically, mismatched content-length error.

Skarlso: I'll write that in the code as well.

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -126,57 +126,53 @@ func (r *RepositoryImpl) getResolver(comp string) (resolve.Resolver, error) {
...
-			Credentials: func(host string) (string, string, error) {
+	authCreds := auth.Credential{}
+	if creds != nil {
+		pass := creds.GetProperty(credentials.ATTR_IDENTITY_TOKEN)
+		if pass == "" {
+			pass = creds.GetProperty(credentials.ATTR_PASSWORD)
+		}
+		authCreds.Username = creds.GetProperty(credentials.ATTR_USERNAME)
+		authCreds.Password = pass
+	}
```
fabianburth, line 137: are we sure the pass in line 131 wouldn't have to be assigned to `authCreds.AccessToken`?

jakobmoellerdev: Sorry to hijack this thread but have to ask: one thing I just noticed after this discussion is that i believe we have an inconsistent property here for token (even in the old code). When do we need ATTR_REGISTRY_TOKEN, ATTR_TOKEN and ATTR_IDENTITY_TOKEN?

Skarlso: 🤷

Skarlso:
Something is not quite right, because I had to cancel the test. That means something is not okay with the added locking.

api/tech/oras/client.go
```diff
@@ -0,0 +1,202 @@
...
+	Client    *auth.Client
+	PlainHTTP bool
+}
+
+type Client struct {
+	Client    *auth.Client
+	PlainHTTP bool
+	Ref       string
+
+	rw sync.Mutex
```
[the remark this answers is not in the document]
jakobmoellerdev: this needs an rw lock, aka RWMutex with Read and Write locks separately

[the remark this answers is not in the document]
jakobmoellerdev: also convention is naming like `fieldToLockMu`

so that would be  `refMu`

jakobmoellerdev:
If it fails in generate I believe this is a concurrency issue in the generator, do we have a global lock ack somewhere?

Skarlso:
@jakobmoellerdev Unsure because it didn't fail, it hang... which is worse. :D

api/tech/oras/client.go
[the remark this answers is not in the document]
fabianburth: on that note, shouldn't ref and plainhttp be private fields?

Skarlso:
Okay, yeah, the mutex in the reader messes things up.

api/tech/oras/client.go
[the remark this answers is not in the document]
Skarlso: the standard is to call it `mu`. I've done that and made the two fields unexported.

Skarlso:
Yay all checks appear to be green.

jakobmoellerdev:
Why would the reader mutex cause a hangup though. That sounds off...

api/tech/oras/client.go
```diff
@@ -0,0 +1,201 @@
...
+	Client    *auth.Client
+	PlainHTTP bool
+}
+
+type Client struct {
+	client    *auth.Client
+	plainHTTP bool
+	ref       string
+	mu        sync.Mutex
+}
```
fabianburth, line 28: On the risk of getting annoying now, I think what @jakobmoellerdev also mentioned in the discussion about this before is, that it would be better to actually use a `sync.RWMutex`. For the actual operations (`Fetch`, `List` and `Push`), a `RLock` would suffice then (I think?).

Skarlso: Fine. :D

api/oci/extensions/repositories/ocireg/repository.go
jakobmoellerdev: For back tracking purposes: Comes from this struct

```
// AuthConfig contains authorization information for connecting to a Registry
type AuthConfig struct {
	Username string `json:"username,omitempty"`
	Password string `json:"password,omitempty"`
	Auth     string `json:"auth,omitempty"`

	// Email is an optional value associated with the username.
	// This field is deprecated and will be removed in a later
	// version of docker.
	Email string `json:"email,omitempty"`

	ServerAddress string `json:"serveraddress,omitempty"`

	// IdentityToken is used to authenticate the user and get
	// an access token for the registry.
	IdentityToken string `json:"identitytoken,omitempty"`

	// RegistryToken is a bearer token to be sent to a registry
	RegistryToken string `json:"registrytoken,omitempty"`
}
```

jakobmoellerdev (approved):

hilmarf:
öhm... why were the integration tests not executed?
https://github.com/open-component-model/ocm/actions/runs/12390047444/job/34584281368?pr=1201

PR #1202, opened 2024-12-18
chore: use the pull request head to run the integration tests

jakobmoellerdev:

This will use the Pull Request Branch Reference and Repository to checkout OCM


jakobmoellerdev:
Force merging this because pull_request_target will not trigger until we have merged.

PR #1205, opened 2024-12-18
fix: context base resolvers used to resolve components

### LINKED ISSUE #1204: Configured resolvers are not used to lookup component names given on the command line
**What happened**:

When configuring resolvers using the ocm config object, they are used to resolve component versions specified
on the command line (here `ocm get component`) without a leading repository part.

But if instead of a component version, only a component name is specified, this does not work.
If a component is specified with repository part, all versions are listed.
 
I would expect that in both cases component version and component names are supported the same way.

**Environment**:

mandelsoft:

Using the ocm config it is possible to configure default resolvers for component versions.
THis is mainly used to implicitly provide lookup repository when following component version references
by recursive operations, like transport and signing.

But they are also used to support the parsing of component version references of the command line.
If configured, the repository part of the component version notation (for example for `ocm get component ....`)
can be omitted and the given name is searched for by using the given resolvers.

Unfortunately this works only for the specification of component versions. But the command line also allows to specify 
component names, only. Typically, then all versions of those components are listed (for the command shown as example).
This is inconsistent, because is doe not for together with resolvers.

This PR introduces the missing case to lookup all versions of a component found via a given component resolver.

Fixes #1204


api/ocm/resolvers/forward.go
```diff
@@ -25,3 +30,62 @@ const (
...
+	}
+	return vp.LookupVersion(version)
+}
+
+func (v *versionResolver) HasVersion(vers string) (bool, error) {
+	cv, err := v.LookupVersion(vers)
+	if err != nil {
+		return false, err
+	}
+	defer cv.Close()
```
jakobmoellerdev, line 89: this is missing an error propagation from close

jakobmoellerdev:
Closing due to inactivity

Issue #1214, opened 2024-12-27
Incompatible change of OCM-Component-Descriptor-Normalisation between OCM-CLI v0.18.0 vs v0.19.0


ccwienk:
# Context

Our delivery-setup encompasses the replication of OCM-Component-Descriptors between multiple OCM-Repositories (dev, staging, ..). We sign our root component-descriptors at the "left-most" OCM-Repository, and validate signatures after replication into "downstream" OCM-Repositories to ensure integrity of replicated contents.

After upgrading from OCM-CLI v0.18.0 to v0.19.0, validation failed for downstream OCM-Repositories (while, interestingly, consistently not failing for the "left-most" OCM-Repository). I _assume_ this deviance might stem from normalisation(s) that happened during downstream replication.

At any rate, it can consistently be shown that `ocm hash componentversions` command outputs different hash-digests for the same componentversion.

# Reproducer

Assuming `ocm-v0.18.0` and `ocm-v0.19.0` are available from PATH, and are copies of OCM-CLI equal to respective version-suffixed, and furthermore assuming OCM-CLI is provided w/ required credentials, the following script will yield different hash-digests:

```
ocm-v0.18.0 \
 ocm hash componentversions \
 --output yaml \
--lookup OCIRegistry::europe-docker.pkg.dev/sap-se-gcp-k8s-delivery/releases-staging-private \
  github.com/gardener/gardener:v1.110.1

# digest: b40006cb7a995be00fdc9c5f85cc4239ee7387288697a9ce2e29482a579bf66d

ocm-v0.19.0 \
 ocm hash componentversions \
 --output yaml \
--lookup OCIRegistry::europe-docker.pkg.dev/sap-se-gcp-k8s-delivery/releases-staging-private \
  github.com/gardener/gardener:v1.110.1

# digest: a2d421e0ede2f34905a1006872e07017d2cd435a0680f3f378a593a5c7ade097
```

Wheareas the following calls (same assumptions as before) will yield equal digests (as expected):

```
ocm-v0.18.0 \
 ocm hash componentversions \
 --output yaml \
--lookup OCIRegistry::europe-docker.pkg.dev/sap-se-gcp-k8s-delivery/releases-private \
  github.com/gardener/gardener:v1.110.1

# digest: b40006cb7a995be00fdc9c5f85cc4239ee7387288697a9ce2e29482a579bf66d

ocm-v0.19.0 \
 ocm hash componentversions \
 --output yaml \
[...]

PR #1218, opened 2025-01-03
feat(normalisation): jsonNormalisation/v3 and fixes to jsonNormalisation/v1 as well as jsonNormalisation/v2

jakobmoellerdev:

Package jsonv3 provides a normalization which is completely based on the abstract (internal) version of the component descriptor and is therefore agnostic of the final serialization format. Signatures using this algorithm can be transferred among different schema versions, as long as is able to handle the complete information using for the normalization. jsonv2 is the predecessor of this version but had internal defaulting logic that is no longer included as part of this normalization. Thus v3 should be preferred over v2. Note that between v2 and v3 differences can occur mainly if the "extra identity" field is not unique, in which case the v2 normalization opinionated on how to differentiate these items. This no longer happens in v3, meaning the component descriptor is normalized as is.

v2 and v1 were adjusted to accomodate the old(but new because forgotten) legacy behavior in legacy.go. Without this, old signatures would not work. This means this should be (at least partially) back-ported to the last minor versioned released without this correction.

This issue fixes https://github.com/open-component-model/ocm/issues/1214 and supercedes https://github.com/open-component-model/ocm/pull/1217 as a better solution longterm (by getting rid of the old normalization) and shortterm (by achieving full backwards compatibility + introducing a simple test case)

Note that this changes the default normalization algorithm to be `jsonNormalisation/v3` instead of `jsonNormalisation/v1` as it is important for users to migrate as soon as possible.


api/ocm/compdesc/norm_test.go
```diff
@@ -284,4 +293,51 @@ var _ = Describe("Normalization", func() {
...
+        imageReference: ghcr.io/bla:2.0.0
+        type: ociRegistry
+    sources: []
+  meta:
+    schemaVersion: v2
+`)))
+		})
+		It("normalizes v1 with extra identity defaulting", func() {
+			n := Must(compdesc.Normalize(cd, compdesc.JsonNormalisationV1))
+			Expect(string(n)).To(StringEqualTrimmedWithContext("[{\"component\":[{\"componentReferences\":[]},{\"name\":\"ocm.software/duplicate-resource/test\"},{\"provider\":\"internal\"},{\"resources\":[[{\"extraIdentity\":[{\"version\":\"1.0.0\"}]},{\"name\":\"image\"},{\"relation\":\"local\"},{\"type\":\"ociImage\"},{\"version\":\"1.0.0\"}],[{\"extraIdentity\":null},{\"name\":\"image\"},{\"relation\":\"local\"},{\"type\":\"ociImage\"},{\"version\":\"2.0.0\"}]]},{\"version\":\"1.0.0\"}]},{\"meta\":[{\"schemaVersion\":\"v2\"}]}]"))
```
Skarlso, line 329: How come the second extraIdentity is `null`? Should that be `2.0.0`?

api/ocm/compdesc/normalizations/legacy/legacy.go
```diff
@@ -0,0 +1,57 @@
...
+	resources := make([]accessors.ElementMeta, len(cd.Resources))
+	for i := range cd.Resources {
+		resources[i] = &cd.Resources[i]
+	}
+	defaultingOfVersionIntoExtraIdentity(resources)
+}
+
+func defaultingOfVersionIntoExtraIdentity(meta []accessors.ElementMeta) {
+	for i := range meta {
+		for j := range meta {
```
Skarlso, line 36: What do you think the frequency of call is here?

If we think this is a point where a lot of times for a lot of resources serialization happens, we should improve this.

Improving it shouldn't be too difficult. We could use a seen := map[string]map[string]struct{} where the second map is the identity digest to compare if we have seen that particular identity for that name we already encountered.

That change would make this a single loop loop. :)

But if you think that's a premature optimization, I don't care. :D

api/ocm/compdesc/norm_test.go
Skarlso: Looking at the implementation now it's clear why it's null.

Skarlso (commented):
One question, otherwise, lovely job Jakob! :)

api/ocm/compdesc/normalizations/legacy/legacy.go
jakobmoellerdev: I can do an identity mapping to get it to O(N), no problem. I think its a good practice even though I dont think it matters all too much (were talking about 100x100 entries in the biggest of components.) Good catch :)

mandelsoft (commented):
I'm not sure, why you created a new normalization v3, which is basically exactly the same as v2, before, and changed v2 to modify the component descriptor.

Normalization should never change anything.

The issue stated, that the problem arises after a transport, therefore it cannot be caused by the normalization, but by the transport handling of the extra identity (as you correctly identified in your comment in the issue).

So, it is caused by the defaulting done in the SetResource and SetSource methods of a CV.  If modify mode is on, it defaults the extra identity according to the new rules. 
This has been intended for the creation of a component version, but this should never happen during a transport.

Therefore, we have to fix the transport behavior, ere. This is only possible by introducing a new modification option for those methods, which instruct the adding to keep the original method data and do not the defaulting. The transport tool then has to give this option. (see #1223)

jakobmoellerdev:
> I'm not sure, why you created a new normalization v3, which is basically exactly the same as v2, before, and changed v2 to modify the component descriptor.
> 
> Normalization should never change anything.

I changed it because your change broke existing logic for normalization. Before, it didnt matter if the component versions contained extra identities or not, now it does. V3 simply makes this break obvious. 

> 
> The issue stated, that the problem arises after a transport, therefore it cannot be caused by the normalization, but by the transport handling of the extra identity (as you correctly identified in your comment in the issue).
> 
> So, it is caused by the defaulting done in the SetResource and SetSource methods of a CV. If modify mode is on, it defaults the extra identity according to the new rules. This has been intended for the creation of a component version, but this should never happen during a transport.

This is true, but at the same time, the normalization is not allowed to change. I still think the change you submitted is relevant, but the normalization must stay the same.

> Therefore, we have to fix the transport behavior, ere. This is only possible by introducing a new modification option for those methods, which instruct the adding to keep the original method data and do not the defaulting. The transport tool then has to give this option. (see #1223)

Also true, but existing (already transferred) component versions have to be verifiable as well, even with the transport having changed the extra identity

Skarlso (approved):

api/ocm/compdesc/normalizations/legacy/legacy.go
jakobmoellerdev:
```

func defaultingOfVersionIntoExtraIdentity(meta []accessors.ElementMeta) {
	// nameToIdentity keeps track of the first occurrence of a given (name+extraIdentity)
	// It maps:
	//   resource name -> (string representation of extra identity -> first index seen)
	nameToIdentity := make(map[string][]int, len(meta))
	// conflicts contains a list of the keys that are conflicting so we can easily get all indices
	conflicts := make(map[string]struct{}, len(meta))

	for i, e := range meta {
		// Stable way to stringify Name+ExtraIdentity.
		key := e.GetName() + e.GetExtraIdentity().String()

		// This is the first time we see this (name + extra identity).
		// Store the index to detect future collisions.
		nameToIdentity[key] = append(nameToIdentity[key], i)

		if len(nameToIdentity[key]) > 1 {
			conflicts[key] = struct{}{}
		}
	}

	for conflict := range conflicts {
		indexes := nameToIdentity[conflict]

		// Overwrite the map entries – we keep the last occurrence as-is
		// and only apply the version to duplicates
		for _, idx := range indexes[:len(indexes)-1] {
			e := meta[idx]

			// A collision is found, so "default the version into extra identity" for the current item.
			eid := e.GetExtraIdentity()
			eid.Set(compdesc.SystemIdentityVersion, e.GetVersion())
			e.GetMeta().SetExtraIdentity(eid)

			Logger.Warn(
				fmt.Sprintf(
					"resource identity duplication was normalized for backwards compatibility, "+
						"to avoid this either specify a unique extra identity per item or switch to %s",
					compdesc.JsonNormalisationV3,
				),
				"name", e.GetName(), "index", idx, "conflicting-indexes", indexes, "extra identity", e.GetExtraIdentity(),
			)
		}
	}
}

```

I will leave this code here in case we need it later on if we have performance issues but its not as readable and I dont think its worth the performance gain, thus resolving. We can easily revisit this later if needed

PR #1224, opened 2025-01-06
fix: use copy uid gid when running toi applications

Skarlso:
Fixes https://github.com/open-component-model/ocm-project/issues/359


hilmarf (approved):

Skarlso:
This is actually making the command fail in a different manner:

```
Error: error copying to / in container: Error response from daemon: getent unable to find entry "1001" in passwd database
```

PR #1229, opened 2025-01-07
fix: do not handle root folder and the . folder

Skarlso:

Related to https://github.com/open-component-model/ocm-project/issues/359

### DIFF
--- a/api/ocm/tools/toi/drivers/docker/driver.go
+++ b/api/ocm/tools/toi/drivers/docker/driver.go
@@ -469,7 +469,7 @@ func generateTar(files map[string]blobaccess.BlobAccess, uid int) (io.ReadCloser
 			dir := path
 			for dir != "/" {
 				dir = unix_path.Dir(dir)
-				if !have[dir] {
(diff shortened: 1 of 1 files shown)


api/ocm/tools/toi/drivers/docker/driver.go
```diff
@@ -469,7 +469,7 @@ func generateTar(files map[string]blobaccess.BlobAccess, uid int) (io.ReadCloser
 			dir := path
 			for dir != "/" {
 				dir = unix_path.Dir(dir)
-				if !have[dir] {
+				if dir != "/" && dir != "." && !have[dir] {
```
frewilhelm, line 472: Isn't `dir != "/"` redundant due to the for-loop on L470?

frewilhelm: Nevermind `unix_path.Dir(dir)` (or rather `path.Dir()`) would cut `/a` to `/`. Thus, it is not covered by the for-loop

frewilhelm (approved):
lgtm

PR #1238, opened 2025-01-09
chore(ci): ensure body is encoded correctly and can never be recognized as secret for release

jakobmoellerdev:

Only passes the body of the release notes and encodes it properly before passing so that we can not run into issues when handling the notes in different steps as an output

https://github.com/open-component-model/ocm/actions/runs/12690118935/job/35370314340#step:13:2 failed

fix https://github.com/open-component-model/ocm-project/issues/363

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,7 +36,7 @@ jobs:
       RELEASE_VERSION_NO_SUFFIX: ${{ needs.release-version.outputs.version_no_suffix }}
       REF: ${{ github.ref }}
     outputs:
-      draft-release-notes: ${{ steps.release-notes.outputs.json }}
+      draft-release-notes-body-base64: ${{ steps.release-notes.outputs.release_notes_body_base64 }}
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


frewilhelm:
I don't understand how base64 encoding can solve this error `2025-01-09T12:56:50.7656452Z ##[error]The template is not valid. .github/workflows/release.yaml (Line: 212, Col: 12): Error reading JToken from JsonReader. Path '', line 0, position 0.
`. 

Do you have any insights on that?

jakobmoellerdev:
Yes. The issue was that the original body was empty and the JSON Decode only failed because that body was empty. The body was empty because of the warning I linked in which GH actions recognized the release JSON as containing secrets. So I changed it so that only the body part and only base64 encoded, gets passed so that the content basically never gets to get filtered by github actions output protection. This will fix the issue of the "empty body".

frewilhelm:
Ok, I understand the issue now.

Why is it not enough to only pass the `body`?

Hiding the `body` using `base64` seems to me like an overkill. Is this really required to solve the current issue or is it an addition for potential future cases, in which a fake/mock secret or text could be trigger the warning?

Skarlso (approved):

jakobmoellerdev:
The base64 is not to hide it. It is to avoid issues related to new lines in the release notes when the body is passed around

Issue #1235, opened 2025-01-09
Bug: `+` in component version is replaced by `.build-`


hilmarf:
**What happened**:

During `ocm transfer` the cv version: `0.0.1-20250108132333+af79499`, gets changed to: `0.0.1-20250108132333.build-af79499`

**What you expected to happen**:

I didn't expect that `+` is replaced by `.build-`. Why is that happening fully automatic?

**How to reproduce it (as minimally and precisely as possible)**:

```yaml
--- # component-constructor.yaml
components:
  - name: docker.io/recmo/docker-hello-tiny
    version: 0.0.1-20250108132333+af79499
    provider:
      name: recmo
    resources:
      - name: docker-hello-tiny
        version: v1.0
        type: ociImage
        access:
          type: ociArtifact
          imageReference: docker.io/recmo/docker-hello-tiny:latest
```

```bash
ocm add cv -c -F .\tmp\ .\component-constructor.yaml
ocm transfer cv .\tmp\ ocm-oci-test.int.repositories.cloud.sap
```

**Error**:

```bash
ocm get cv ocm-oci-test.int.repositories.cloud.sap//docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499 -o yaml
```

closing session: unable to close: closing docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499: unable to unref last: unable to cleanup component version  docker.io/recmo/docker-hello-tiny/0.0.1-20250108132333.build-af79499 while unref last: closing component version docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499: check failed: component version "0.0.1-20250108132333+af79499" is invalid

**Environment**:

ikhandamirov:
I suppose, the '+' character is not allowed in the version as per specification:
```
Image tags consist of lowercase and uppercase letters, digits, underscores (_), periods (.), and dashes (-). It can be up to 128 characters long. And must follow the next regex pattern: [a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}
```
and therefore getting replaced. See also: https://kubernetes.io/docs/concepts/containers/images/

ikhandamirov:
Same in the OCI spec:
```
Throughout this document, <reference> as a tag MUST be at most 128 characters in length and MUST match the following regular expression:

[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}
```

[...]

PR #1243, opened 2025-01-10
fix: unable to close, unable to unref last: unable to cleanup component version, while unref last: closing component version

hilmarf:

When calling ocm-cli with oci-tag-version like `0.0.1-20250108132333.build-af79499` e.g.

```bash
ocm get cv ocm-oci-test.int.repositories.cloud.sap//docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499
```

It fails during the [finalize step](https://github.com/open-component-model/ocm/blob/releases/v0.19/api/utils/refmgmt/refcloser.go#L196), when [doing an update](https://github.com/open-component-model/ocm/blob/releases/v0.19/api/ocm/extensions/repositories/genericocireg/componentversion.go#L171), where [another round of version comparison](https://github.com/open-component-model/ocm/blob/releases/v0.19/api/ocm/extensions/repositories/genericocireg/componentversion.go#L99) happens.

Fixes: https://github.com/open-component-model/ocm/issues/1235

Related: https://github.com/open-component-model/ocm/pull/355


api/ocm/extensions/repositories/genericocireg/component.go
```diff
@@ -128,6 +128,10 @@ func (c *componentAccessImpl) HasVersion(vers string) (bool, error) {
 }
 
 func (c *componentAccessImpl) LookupVersion(version string) (*repocpi.ComponentVersionAccessInfo, error) {
+	// LookupVersion '0.0.1-20250108132333.build-af79499' would fail with:
+	// unable to unref last: unable to cleanup component version [%v] while unref last: closing component version [%v]: check failed: component version [%v] is invalid
```
Skarlso, line 132: Does the actual error contain `[%v]` instead of the version? Just curious. :D

hilmarf: no - the actual error is:

```text
closing session: unable to close: closing docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499: unable to unref last: unable to cleanup component version docker.io/recmo/docker-hello-tiny/0.0.1-20250108132333.build-af79499 while unref last: closing component version docker.io/recmo/docker-hello-tiny:0.0.1-20250108132333.build-af79499: check failed: component version "0.0.1-20250108132333+af79499" is invalid
```

Skarlso (approved):

api/ocm/extensions/repositories/genericocireg/component.go
mandelsoft: The problem is, that a version with this `.build-` is not a valid version anymore for CVs. So, the error is basically correct, but it comes far too late.

It must be assured that the version name given at the API matches the version name stored in the component descriptor, otherwise the complete structure would be corrupted.

So, such a version should be rejected at all. Accepting two versions, the CV version and a technical versions used at the backend, as valid CV versions would be problematic. But there is no reason why to accept the technical version, also. (The `ListVersions` call correctly maps the technical versions found as OCI tag back to OCM versions).

So, there should be an error if the technical version is given, and it should be reported early.
May be you could add the following check to `LookupVersion` in  api/ocm/cpi/repocpi/bridge_c.go  Line 102
```
func (b *componentAccessBridge) LookupVersion(version string) (cpi.ComponentVersionAccess, error) {
	i, err := b.impl.LookupVersion(version)
	if err != nil {
		return nil, err
	}
	if i == nil || i.Impl == nil {
		return nil, errors.ErrInvalid("component implementation behaviour", "LookupVersion")
	}
	if i.Impl.GetDescriptor().Version != version {
		i.Impl.Close()
		return nil, errors.ErrInvalid("component version mismatch")
	}
	return NewComponentVersionAccess(b.GetName(), version, i.Impl, i.Lazy, i.Persistent, !compositionmodeattr.Get(b.GetContext()))
}
```

mandelsoft (dismissed):
The technical version used at the backend should not be usable as OCM version (see my comment above). I would strongly recommend to provide an error instead of introducing this ambiguity.

jakobmoellerdev (approved):
With the warning in place, we should be good to go not to break anyone while at the same time preparing to forbid this alltogether like suggested. For now, we need to avoid completely breaking here.

Issue #1242, opened 2025-01-10
<User Story> - Add deprecation warning for CA


morri-son:
**User Story**: Add deprecation warning to all code and documentation related to component archive (CA)

**Description**:
As an end user of the OCM CLI and the OCM library I want to see a deprecation warning for all interactions and in all documents related to the component archive (CA) that clearly states the end of support for the CA. The deprecation starts with the announcement (already announced [here](https://kubernetes.slack.com/archives/C05UWBE8R1D/p1734357630853489)) and ends on 2025-06-30.

**Epic link**:

- open-component-model/ocm#1105

**Acceptance criteria**:

- [x] all actions within the library using CAs have an additional deprecation warning in their log/stdout
- [x] all documents, especially the OCM CLI help, referring to a CA have a deprecation information

**Related issues**:

hilmarf:
what about: "[Download component versions from an OCM repository. The result is stored in component archives.](https://ocm.software/docs/cli-reference/download/componentversions/#description)" ???

where do we store component versions when not using component archives?

morri-son:
> what about: "[Download component versions from an OCM repository. The result is stored in component archives.](https://ocm.software/docs/cli-reference/download/componentversions/#description)" ???
> 
> where do we store component versions when not using component archives?

we plan to use CTF archives to store OCM components

fabianburth:
For now, we decided that the documentation might be required for users wanting to migrate away from component archives. Thus, we want to keep the documentation.

PR #1244, opened 2025-01-10
fix: Add deprecation warning for CA

hilmarf:

we don't want to support component archives (CA) anymore

fixes: https://github.com/open-component-model/ocm/issues/1242

potential follow up: https://github.com/open-component-model/ocm/issues/1254


ikhandamirov (commented):
Please see my other comments.

ikhandamirov:
One more place to deprecate:
https://github.com/open-component-model/ocm/blob/0296603c824b3883e81a3f6d665937f5fbe59caf/cmds/ocm/commands/ocmcmds/common/options/fileoption/option.go#L22

hilmarf:
> If I execute, for example, `ocm create ca`, I get the warning:
> 
> ```
> Command "componentarchive" is deprecated, Deprecated - use CommonTransportFormat instead
> ```
> 
> But `componentarchive` is ~not a command~. Unfortunately, I didn't find at first glance, how to correct the message.
> 
> Edit: it is a command. Though the message looks strange, we'll likely have to live with it.

Cobra-commands do actually have a field, where you can write a string to, marking commands as deprecated. See: https://github.com/hilmarf/ocm/blob/deprecate/ComponentArchive/cmds/ocm/commands/ocmcmds/componentarchive/cmd.go#L24
There are in total three places:
image

hilmarf:
> Was it really part of #1244, to directly remove all mentions of component archive from the reference? I thought, for now we only wanted to put a disclaimer with a warning.
> 
> But I'm not against removing it. I find it a bit inconsistent, but, on the other hand, maybe it is a good motivation for existing user to start thinking of switching to CTF.

You're right, it's not written as requirement to remove all mentions of component archive. But I don't like to read through anyhow too large documentation and then need to let my brain filter out everything which is deprecated.

@morri-son what's your take on that? Shall we still generate all the documentation around CAs or not?

hilmarf:
> One more place to deprecate:
> 
> https://github.com/open-component-model/ocm/blob/0296603c824b3883e81a3f6d665937f5fbe59caf/cmds/ocm/commands/ocmcmds/common/options/fileoption/option.go#L22

thanks - good catch! I obviously missed to search for `component-archive` - for whatever reason it's written here with '-'!?!?

ikhandamirov:
@hilmarf, Commands `ocm add sources` and `ocm add resources` are specific to component archives:
```
So far only component archives are supported as target.
```
So far those commands are not covered by this PR, if I see it correctly, right? There may be also other commands like this, which I'm not aware of.

hilmarf:
> @hilmarf, Commands `ocm add sources` and `ocm add resources` are specific to component archives:
> 
> ```
> So far only component archives are supported as target.
> ```
> 
> So far those commands are not covered by this PR, if I see it correctly, right? There may be also other commands like this, which I'm not aware of.

yes, that's right - https://github.com/open-component-model/ocm/issues/1254 is the follow up task to be done... or we are faster with the rewrite and skip it ;-)

hilmarf:
> A separate question is, if we remove all information about component archives, do we already have enough documentation on how to achieve the same result with CTF?

I don't know. I'm not using CAs.

morri-son:
> A separate question is, if we remove all information about component archives, do we already have enough documentation on how to achieve the same result with CTF?

as mentioned above there are some commands that only work with CAs, but personally I also never used CAs, but only CTFs and think there is enough documentation on using CTFs + component constructors. Only in case someone manually creates components step by step using resource files, he may stumble into this issue. Looking at the documentation these cases are more to explain how to build a component and its different artifacts step by step, not a use case for pipelines.

ikhandamirov:
@morri-son , please check https://github.com/open-component-model/ocm/pull/1244#issuecomment-2598323623

As you see, @hilmarf suggests to remove documentation for `ocm create ca` and `ocm transfer ca` as part of this PR.
1. Do we want to do this, or do we for now only want to add a deprecation warning to the documentation?
2. Commands like `ocm add sources` and `ocm add resources` are also specific to component archives. Do we want to remove their documentation as well? Or do we want to tag them with `FIXME` and adapt to CTF with #1254?

morri-son:
> @morri-son , please check [#1244 (comment)](https://github.com/open-component-model/ocm/pull/1244#issuecomment-2598323623)
> 
> As you see, @hilmarf suggests to remove documentation for `ocm create ca` and `ocm transfer ca` as part of this PR.
> 
> 1. Do we want to do this, or do we for now only want to add a deprecation warning to the documentation?
> 2. Commands like `ocm add sources` and `ocm add resources` are also specific to component archives. Do we want to remove their documentation as well? Or do we want to tag them with `FIXME` and adapt to CTF with [<User Story> - Replace CA usage in resources + download #1254](https://github.com/open-component-model/ocm/issues/1254)?

I can only share my point of view :-) I do not see a benefit in mentioning CAs anywhere in the docs anymore when we're about to discontinue support for them soon. We will keep it in the code for own purposes, e.g. debugging, but we want to shift end users to CTFs already right now.

Therefore I would also remove `ocm add resources` and `ocm add sources` (and potential other commands forcing the use of CAs)  from the documentation / website completely instead of having a warning in the docs. Throwing warnings and give recommendations when people use such commands (because they maybe use an older copied version of the docs, which is already highly hypothetical) is the right way, but in the docs I would try to remove all occurrences of the component archives wherever consistently possible.

Skarlso:
My only comment is that the documentation should contain what to use instead! I know that the slack message contains some hints, I don't want to open slack when I'm reading documentation about a thing I'm about to use. I want to read it right there in the deprecation notice.

hilmarf:
> My only comment is that the documentation should contain what to use instead! I know that the slack message contains some hints, I don't want to open slack when I'm reading documentation about a thing I'm about to use. I want to read it right there in the deprecation notice.

Do you have any suggestion what the user should use instead? And how? I'm happy to add the proper text.

jakobmoellerdev:
Everywhere CAs are outdated, CTFs are the appropriate replacement. So whenever having the deprecation warning, there can be sentence on how to replicate the flow with CTFs

hilmarf:
> Everywhere CAs are outdated, CTFs are the appropriate replacement. So whenever having the deprecation warning, there can be sentence on how to replicate the flow with CTFs

For most cases that's a pretty simple: `(DEPRECATED) - Please use commontransportarchive instead`.
But I don't know what to write for: `ocm add resource` and `ocm download componentversions`.

hilmarf:
> I have no idea. I'm not versed enough in OCM. That's why I was missing such context. :D

@Skarlso here you go: https://github.com/open-component-model/ocm/pull/1244/commits/7adebf0304d47b3a3f762b51b10e648339fe9079

Skarlso (approved):

PR #1252, opened 2025-01-14
chore(format): executed make format on EVERYTHING (incl. tests)

hilmarf:
everytime, I'm executing `make format`, those files get changed, even when they have nothing to do with the actual feature


frewilhelm:
Shouldn't we have an action to prevent code to be merged that is not correctly formatted? :O

frewilhelm (approved):

hilmarf:
> Shouldn't we have an action to prevent code to be merged that is not correctly formatted? :O

I believe we have that check, but it checks only on productive code and not on tests :-/

PR #1255, opened 2025-01-14
chore: change epic template

morri-son:
add some additional points to epic template


frewilhelm:
IMO the `ocm` repository is not the correct repository to store this template. I would propose the project or .github repository

morri-son:
> IMO the `ocm` repository is not the correct repository to store this template. I would propose the project or .github repository

@frewilhelm Created PR https://github.com/open-component-model/.github/pull/17 and copied over everything from the OCM repo. Once merged, we can remove all distinct templates from the other repos, as long as we don't want to overwrite the central templates with something repo-specific.

hilmarf:
@morri-son https://github.com/open-component-model/.github/pull/17 is merged, what about this one?

morri-son:
closed as contained in https://github.com/open-component-model/.github/pull/17

PR #1329, opened 2025-02-27
chore: switch blackduck sca to pull_request_target

jakobmoellerdev:

makes sure that we run with pull request target. this is still safe because in ocm all external (untrusted contributors) must be approved before running the PR


morri-son (approved):
lgtm

PR #1344, opened 2025-03-10
feat: optimising ocm transfer approach with concurrent worker pool

us-2608:
This PR is for optimising the transfer of sources and resources using concurrency , which helped in transfer time reduction


jakobmoellerdev:
Hi there, thanks for your contribution. Please take a look at the test scenarios and make sure that we are not regressing.

Furthermore:
- Some of the log messages look very strange: https://github.com/open-component-model/ocm/actions/runs/13759312005/job/38471923676?pr=1344#step:22:562 `tasktasktask...adding component version...`
- It looks like some of the updates to the component descriptor accesses are not issued correctly, eg. imageReferences do not get updated: https://github.com/open-component-model/ocm/actions/runs/13759312005/job/38471923676?pr=1344#step:22:743

Issue #1362, opened 2025-03-18
ocm transfer components --recursive is not working as expected


avinashreddynarra89:
**Context**  
I am trying to transfer componentversions using ocm transfer command but it is not recursively searching for sub components and giving error unable to access the meta data of the manifest version.As the metadata actually exists.It works fine if the version is v0.12.0 but from v0.13.0 and above issue occurs.

Issue is happening after this commit
https://github.com/open-component-model/ocm/pull/784/files

And also after this commit it is trying to get manifest version of directory instead of going inside  directories and search for manifest version of particular component.

**Version**  
Above v0.12.0

**To Reproduce**  
Its happening from the versions v0.13.0 and above

**Actual behavior**  
It is not recursively searching for sub components manifest version

**Expected behavior**  
It should recursively search for sub components manifest version

**Screenshots (optional)**  

**Environment**  
- [ ] windows
- [ ] linux
- [x] mac

**Additional Comments**  
Any observation, gut feeling or other information you think help the community to troubleshoot the problem?

avinashreddynarra89:
@frewilhelm can you please help on this issue as it is giving us issues which needs to be resolved asap

frewilhelm:
Hi, @avinashreddynarra89 we first need to validate the issue. We will keep you updated

fabianburth:
Hey, @avinashreddynarra89 we need to support here. 

Could you provide component descriptors with which you encounter this issue? And could you provide some information with regards to how you encountered the bug? Are you still using version v0.12.0 and currently tried to upgrade?

avinashreddynarra89:
@fabianburth  existing target version: closing <url>/kubernetes-private/landscape-clusters-setup:0.31.0: unable to unref last: unable to cleanup component version  <url>/kubernetes-private/landscape-clusters-setup/0.31.0 while unref last: closing component version <url>/kubernetes-private/landscape-clusters-setup:0.31.0: cannot access meta data manifest version: failed to resolve manifest "<url>/component-descriptors:meta": HEAD "https://<url>/component-descriptors/manifests/meta": response status code 400: Bad Request

It is trying to get manifests for directory instead of not going inside directory and get component manifest version


for work around we are still using v0.12.0 but we need to get the latest version working

fabianburth:
Could you also provide the exact command you used and the component descriptors?

avinashreddynarra89:
@fabianburth 
[...]

PR #1367, opened 2025-03-20
fix: make sure that other resolve requests such as 400 are also returned as unknown errors so that OCM recognizes them

jakobmoellerdev:

Makes sure resolution of artifacts resolves as Unknown Error for OCM if its something like a 400

fix https://github.com/open-component-model/ocm/issues/1362

### DIFF
--- a/api/oci/extensions/repositories/ocireg/namespace.go
+++ b/api/oci/extensions/repositories/ocireg/namespace.go
@@ -147,7 +147,8 @@ func (n *NamespaceContainer) GetArtifact(i support.NamespaceAccessImpl, vers str
 		if errdefs.IsNotFound(err) {
 			return nil, errors.ErrNotFound(cpi.KIND_OCIARTIFACT, ref, n.impl.GetNamespace())
 		}
-		return nil, err
+
+		return nil, errors.ErrUnknown(cpi.KIND_OCIARTIFACT, err.Error())
 	}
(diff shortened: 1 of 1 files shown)


ikhandamirov:
Do I understand correctly that the fix hides the error? But we don't know why it happened and why it was not reproducible? The root cause seems to be different behavior of Docker registry compared to Artifactory?

jakobmoellerdev:
We do know why it happened and how. The pub sub check happens all the time, every time. It is part of the OCM pub sub negotiation. The error occurs because an ORAS error on the manifest fetch for the discovery of said pub sub config was not wrapped correctly (400 was not wrapped as unknown). Indeed it is part of the contract of the pub sub interface that an unknown error is simply ignored. 

Long term we can think of disabling pubsub by default.

jakobmoellerdev:
See https://github.com/open-component-model/ocm/blob/c76a2ed6ff3402451bf5a741ca8e1676e8381ab2/api/ocm/extensions/pubsub/providers/ocireg/provider.go#L48

ikhandamirov:
Yes, exactly this `if` (also linked in my first message) is worth a more in-depth discussion. I'm approving, of course, as we saw the fix in action.

ikhandamirov (approved):

Skarlso (approved):

Issue #1366, opened 2025-03-20
Authorization failure when using docker config and Docker wincred credential helper


KevinLubbersen:
**Context**  
Since version v0.20.0 of OCM authorization to Azure Container Registry fails when using Docker configuration with the wincred credential helper.

**Version**  
v0.20.0
v0.20.1
v0.21.0

**To Reproduce**  
1. Login to an Azure Container Regstry using Azure CLI (e.g. `az acr login -n my-registry`)
2. List component versions using ocm cli (e.g. `ocm list componentversion --repo OCIRegistry::my-registry.azurecr.io company.com/my-component`
3. Notice the authorization failure

**Actual behavior**  
Authorization fails with OCM v0.20.0 or later.

**Expected behavior**  
ocm commands should authn/authz successfully, as is the case when using OCM v0.19.x or prior.

**Environment**  
- [x] windows
- [ ] linux
- [x] mac

Skarlso:
Reproduced the Azure bug.

```
❯ ./bin/ocm list componentversion --repo OCIRegistry::skarlso.azurecr.io company.com/my-component                                                                                                                                       (base)
Error: error processing "company.com/my-component": failed to list tags: GET "https://skarlso.azurecr.io/v2/component-descriptors/company.com/my-component/tags/list": GET "https://skarlso.azurecr.io/oauth2/token?scope=repository%3Acomponent-descriptors%2Fcompany.com%2Fmy-component%3Ametadata_read%2Cpull&service=skarlso.azurecr.io": response status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: 828def6a-27d3-47bd-9e55-92ebd03f6293
```

Skarlso:
Yes, it's working on 0.19. So this is definitely ORAS.

PR #1370, opened 2025-03-21
fix: set the refresh token to password always

Skarlso:

Fixes https://github.com/open-component-model/ocm/issues/1366


Skarlso:
I need a different way.

Skarlso:
I'll probably need some kind of dynamic store.

Skarlso:
There might be a better workaround for this here: https://github.com/Azure/acr/blob/main/docs/AAD-OAuth.md#authenticating-docker-with-an-acr-refresh-token

Skarlso:
Current thing might work. Going to run some further tests against other repos and Azure.

hilmarf (approved):

Skarlso:
This isn't working yet actually!

Skarlso:
Still have some trouble... I'm figuring out the right thing to use.

hilmarf (approved):

PR #1380, opened 2025-03-24
feat: add context to support cancel to tool ops

### LINKED ISSUE #1346: Question: how can I cancel a TransferVersion when using OCM as a module
I am using OCM as a module in my own application. One of the API's I am using is the `TransferVersion` function.
I am using it to transfer component versions from one repository to another including all resources. Some of these resources are quite large and will take some time to transfer.

I would like to be able to abort the action but I cannot find a way to accomplish this.

Is this possible and if so, could you point me in the right direction?

mandelsoft:

This PR changes the  internal tool operations for signing and transfer to use a `context.Context`.
This context can be used to cancel the operations. The check is always done before a new component version is
handled. This means, if cancelled, the actual component version handled after a cancel will always be finished before
the operation is cancelled.

The consumer facing operations are  added to provide versions accepting a context.

Fixes #1346


jakobmoellerdev:
Do I understand correctly that within one component version transfer, there is no cancellation, but between multiple component versions there is? Meaning that large resource transfers within one cv are not cancellable but multiple component versions within a transfer are?

mandelsoft:
The proposed solution checks for cancellation just before starting the handling of a new/next CV, but not during the transfer of content of a CV. This means, after starting the transfer of a CV all resources are transferred,, and if there are nested CVs with the
call of operation for those CVs a cancellation is checked, again. 
This might be cause a longer cancellation time for large CVs, but deep transfers can be cancelled.

If this is a problem, it would be possible to introduce more fine grained checkpoints, but I think we could start this way.

jakobmoellerdev:
In the issue it was specifically asked for cancellations in between large resources, how would you want to tackle that? Still I think we should be good to accept this change as well.

mandelsoft:
I've refined the cancellation check to be able to cancel among resource handlings, also.
If we want to cancel the particular resource handling, we would have to change the API for Uploaders and the 
resource API on a Component version.

jakobmoellerdev:
I am okay with merging this after you adjusted the generation. I think that in between two resources the transfer can still get stuck though (as you said this would change the resource api). I think for now this change is a good start.

jakobmoellerdev (approved):

PR #1406, opened 2025-04-04
fix: make sure that ref is an option in the gitHub AccessMethod

jakobmoellerdev:

previously ref was lost in the access method. This adds it back in...


frewilhelm (approved):
lgtm

PR #1412, opened 2025-04-09
feat: configurable preference for relative access creation for uploader

### LINKED ISSUE #1410: relative accesses: using absolute access types for transferred-by-value artifacts is problematic
### Scenario

- A K8s cluster runs an own OCI registry.
- All artifacts of a component version should be made available in that cluster-internal OCI registry (no references to external OCI registries)
- Consumption of the component version artifacts happens from within the cluster using the cluster-local network only (all external network access is blocked).

The component version is transferred into the cluster-internal registry using _port-forwarding_:

```
$ ocm transfer ctf --copy-resources component-1/ http://localhost:9999/local/components
```

### Problem

The artifact consumption from within the cluster fails.

Example using OCM CLI:

```
$ ocm download resources http://cluster-registry.svc/local/components//example.com/component-1 image--busybox -O busybox.tar.gz
Error: resource image--busybox/v1.36.1: unable to read access: failed to resolve manifest "localhost:9999/local/components/library/busybox@sha256:ec15a5bd53cf9507beb851574654669e778a9735f8e605e0ee3d71fd07debbe1": Head "http://localhost:9999/v2/local/components/library/busybox/manifests/sha256:ec15a5bd53cf9507beb851574654669e778a9735f8e605e0ee3d71fd07debbe1": dial tcp [::1]:9999: connect: connection refused
```

It tries to get the artifact (`image--busybox`) using the registry host name + port `localhost:9999` that was used when uploading the component version. But this is not valid in the consumer's environment.

The component version descriptor in the registry looks like this:

```yaml
meta:
  schemaVersion: v2
component:
  componentReferences: []
  creationTime: "2025-04-07T13:01:16Z"
  name: example.com/component-1
  provider: example.com
  repositoryContexts:
  - baseUrl: http://localhost:9999
    componentNameMapping: urlPath
    subPath: local/components
    type: OCIRegistry
  resources:
  - access:
      imageReference: http://localhost:9999/local/components/library/busybox:1.36.1@sha256:ec15a5bd53cf9507beb851574654669e778a9735f8e605e0ee3d71fd07debbe1
      type: ociArtifact
    digest:
      hashAlgorithm: SHA-256
      normalisationAlgorithm: ociArtifactDigest/v1
      value: ec15a5bd53cf9507beb851574654669e778a9735f8e605e0ee3d71fd07debbe1
    name: image--busybox
    relation: external
    type: ociImage
    version: v1.36.1
  sources: []
  version: 0.1.0
```

`component.resources[0].access` is an _absolute_ reference, although the artifact is stored with the component descriptor in the same OCM repository.

### Proposed Solution

For all artifacts contained in the same OCM repository as the component version descriptor _relative_ access types should be used (relative to OCM repo location of component version descriptor). This avoids storing the URL of the OCM repository, and thus makes the descriptor portable.

### Further Information

The described problem arises in all scenarios where the OCM repo URL during _uploads_ of component versions differs from the OCM repo URL used by consumers, for example:

- Consumers access an OCI repository through a reverse proxy
- OCI repository content  is generically replicated to other instances, which are used by consumers

mandelsoft:

By default the OCI uploader used to handle the implicit OCI uploads for OCI based OCM repositories
creates an absolute access method.

A new global attribute `preferrelativeaccess` (bool)  can be used now to switch this behavior
to prefer a relative access method.

With the new config object `local.oci.uploader.config.ocm.software` it is possible to configure
host/port combinations for which the relative access should be preferred. (For example `localhost`).

```
type: local.oci.uploader.config.ocm.software
preferRelativeAccess: true
repositories:
- localhost
- other:5000
```

Fixes #1410


api/config/utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	cfgcpi "ocm.software/ocm/api/config/cpi"
+)
+
+// GetConfigured applies config objects of a config context
+// to a configuration struct of type T.
+// A pointer to the configured struct is returned.
+// Attention: T must be a struct type.
+func GetConfigured[T any](ctxp ContextProvider) (*T, error) {
```
jakobmoellerdev, line 11: There is exactly one usage of this method, so I would like to keep it an internal helper

api/oci/internal/repotypes.go
```diff
@@ -20,11 +20,15 @@ type IntermediateRepositorySpecAspect interface {
 	IsIntermediate() bool
 }
 
+type UniformRepositorySpecProvider interface {
+	UniformRepositorySpec() *UniformRepositorySpec
+}
+
 type RepositorySpec interface {
 	runtime.VersionedTypedObject
+	UniformRepositorySpecProvider
```
jakobmoellerdev, line 29: Why is every Repository Spec now a Uniform Spec Provider?

api/ocm/extensions/attrs/preferrelativeattr/attr.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+
+	"ocm.software/ocm/api/datacontext"
+	"ocm.software/ocm/api/utils/runtime"
+)
+
+const (
+	ATTR_KEY   = "github.com/mandelsoft/ocm/preferrelativeaccess"
```
jakobmoellerdev, line 11: This attribute key should be changed.
```suggestion
	ATTR_KEY   = "ocm.software/ocm/oci/" + ATTR_SHORT
```

docs/reference/ocm_configfile.md
```diff
@@ -168,6 +168,27 @@ The following configuration types are supported:
 
   At least the given values must be present in the certificate
   to be accepted for a successful signature validation.
+- <code>local.oci.uploader.config.ocm.software</code>
+  The config type <code>local.oci.uploader.config.ocm.software</code> can be used to set some
```
jakobmoellerdev, line 172: This should become part of the generic oci uploader configuration, e.g. `oci.uploader.config.ocm.software` because there is no semantic relationship between a "local.oci.uploader" and the relative access setup. (localhost is only used in the given examples and in tests but it could be any repository).

jakobmoellerdev (changes requested):
In general, relativeOciAccesses are not nicely specified and theyre already not part of the spec anymore. Nevertheless I recognize the value this gives for local debugging.

For me this PR is acceptable, but only if we mark the type as experimental somehow. I think relativeOciAccesses should be avoided because most tools will only know how to deal with ociArtifact / ociImage. Introducing yet another type that is only resolvable if a context is known is very dangerous.

jakobmoellerdev (approved):

PR #1430, opened 2025-04-29
chore: pin `mdref` to v0.5.1

hilmarf:
@mandelsoft created a new tag: https://github.com/mandelsoft/mdref/releases/tag/v0.5.1 which doesn't result in sha-sum mismatches.

We should not use the ugly workaround with the `GONOSUMDB` setting.

fixes: https://github.com/mandelsoft/mdref/issues/1

### DIFF
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -76,8 +76,7 @@ endif
 
 .PHONY: mdref
 mdref:
-	# https://github.com/mandelsoft/mdref/issues/1 as workaround set GONOSUMDB
-	GONOSUMDB='github.com/mandelsoft/mdref' go install -v github.com/mandelsoft/mdref@v0.5.0
+	go install -v github.com/mandelsoft/mdref@v0.5.1
 
 Linux_jq:
 	$(info -> jq is missing)


frewilhelm (commented):
This pin will be recognised by dependabot, right?

hilmarf:
> This pin will be recognised by dependabot, right?

nope, I don't think so. I haven't seen any samples where Dependabot automatically checked Makefiles... That would require some custom dependabot config.

jakobmoellerdev (approved):

hack/Makefile
```diff
@@ -76,8 +76,7 @@ endif
 
 .PHONY: mdref
 mdref:
-	# https://github.com/mandelsoft/mdref/issues/1 as workaround set GONOSUMDB
-	GONOSUMDB='github.com/mandelsoft/mdref' go install -v github.com/mandelsoft/mdref@v0.5.0
+	go install -v github.com/mandelsoft/mdref@v0.5.1
```
ikhandamirov, line 79: It used to pin `@master` before failure. Do we want to stick to a specific version now?

PR #1433, opened 2025-04-29
chore: cleanup migration + duplicate

mandelsoft:

- Cleanup some artifacts from the migration
- deprecate duplication functions in api/utils

- #### Which issue(s) this PR fixes


jakobmoellerdev (commented):
Whats the reason for deprecating the OCM internal utils in favor of mandelsoft? to me this is undesired during a cleanup and duplication removal that is cross-package. We should only upgrade usages that already exist.

mandelsoft:
The functions are duplicates, because during the migration not all occurrences have been adapted.
One instance should be sufficient. An we decided mid of last year to extract general functionality not bound to the ocm functionality.

The alternative is to migrate back this completely to the ocm library. But we should have only one source of truth for those functions.

frewilhelm:
> But we should have only one source of truth for those functions.

In theory, I agree. But using libraries from private contributors or without community support is risky. What happens if the contributor no longer has access to the repositories or makes them private? That can lead to much bigger problems.

Skarlso:
>The alternative is to migrate back this completely to the ocm library.

Let's do that because of the above, what Frederic mentioned. :)

PR #1460, opened 2025-05-16
fix(revert): make sure that ref is an option in the gitHub AccessMethod (#1406)

jakobmoellerdev:
This reverts commit 43632aa3.

We have too many users reporting issues with the ref+commit combination as this was not spec compliant. We need to revert this change asap to stop breaking existing component version flows


jakobmoellerdev:
verify markdown had a regression in a recent bump so im gonna prompt to ignore this check

frewilhelm (approved):

PR #1463, opened 2025-05-16
chore: remove winget

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/499


Skarlso:
@morri-son I imagine we must also somehow unload the existing packages?

morri-son:
> @morri-son I imagine we must also somehow unload the existing packages?

yes, we should do that. @hilmarf , I guess you still have the token used for widget somewhere locally? Could you please remove the packages?

hilmarf:
> @morri-son I imagine we must also somehow unload the existing packages?

why?

hilmarf:
> yes, we should do that. @hilmarf , I guess you still have the token used for widget somewhere locally? Could you please remove the packages?

1. I don't understand what is wrong with the existing packages?
2. Just file a PR which removes the complete folder: https://github.com/microsoft/winget-pkgs/tree/master/manifests/o/Open-Component-Model

morri-son:
> > @morri-son I imagine we must also somehow unload the existing packages?
> 
> why?

To avoid releases that are not maintained anymore. I guess we can keep them right now, but somewhen we should clean them up.

hilmarf:
https://github.com/microsoft/winget-pkgs/discussions/16767

jakobmoellerdev:
Guys, please lets not go overboard. Whats the reason to remove existing packages also for the userbase? IMO this creates not only an issue for existing users (granted there may be not that many which is why we decided to discontinue winget in the first place), but furthermore it creates unneeded cleanup work.

WDYT?

morri-son:
> Guys, please lets not go overboard. Whats the reason to remove existing packages also for the userbase? IMO this creates not only an issue for existing users (granted there may be not that many which is why we decided to discontinue winget in the first place), but furthermore it creates unneeded cleanup work.
> 
> WDYT?

@as mentioned, we can keep the current packages and just remove the upload of new ones. But we should have a plan to remove the old ones somewhen as we move on without winger support. I just have the fear that we forget them :-)

Skarlso:
Honestly, I was just asking. :D I would have been content with a no. :D

Issue #1469, opened 2025-05-20
OCM Resource Downloader with "signed" OCI Helm Charts returns incorrect OCI layer


sk31337:
**Context**  
I have the following OCM `component-constructor.yaml` which references two "signed" Helm Charts from an OCI registry:

```yaml
     - name: oci-helm-chart-mariadb
        type: helmChart
        version: "3.0.3"
        access:
          type: ociArtifact          
          imageReference: oci://registry.opencode.de/bmi/opendesk/components/platform-development/charts/opendesk-mariadb/mariadb:3.0.3
      
      - name: oci-helm-chart-postgresql
        type: helmChart
        version: "2.1.2"
        access:
          type: ociArtifact          
          imageReference: oci://registry.opencode.de/bmi/opendesk/components/platform-development/charts/opendesk-postgresql/postgresql:2.1.2
```

I build & push with: `ocm add components --create --file ./ocm/cft ./ocm/component-constructor.yaml && ocm transfer ctf --overwrite ./ocm/cft mcp-blueprints.common.$REPOSITORY/ocm --enforce`

As you can see via `oras` there are multiple layers of the OCI Helm Charts with a different order:

```bash
# Layer that contains Helm Chart is located in layer 0
➜  git:(main) ✗ oras pull registry.opencode.de/bmi/opendesk/components/platform-development/charts/opendesk-mariadb/mariadb:3.0.3 --output . 
✓ Skipped     application/vnd.cncf.helm.chart.content.v1.tar+gzip                                                                                                   26.6/26.6 KB 100.00%     0s
  └─ sha256:5dd8d63a13140a808960f01b1e9f03da7152a815e6b484339bfa76453d8ab361
✓ Skipped     application/vnd.cncf.helm.chart.provenance.v1.prov                                                                                                    1.33/1.33 KB 100.00%     0s
  └─ sha256:ace7860124acf45595f2654708d130cb304fd7ffd78f1541b4a21d7c66cfdc48
✓ Skipped     application/vnd.cncf.helm.config.v1+json                                                                                                                439/439  B 100.00%     0s
  └─ sha256:09a658e188bf159066cd41b9650f0d65f1ba44179d60430cf24266875b90cc6c
✓ Pulled      application/vnd.oci.image.manifest.v1+json                                                                                                              856/856  B 100.00%     0s
  └─ sha256:a8df0f96133fb1759defdf520b93d04b3fb73f374ce10761bad1eff75089ae4e
[...]

PR #1470, opened 2025-05-20
fix: allow spec compliant, but unusually ordered HELM Chart OCI Artifacts in HELM OCI Artifact downloader

jakobmoellerdev:

This allows arbitrarily ordered, but spec (https://github.com/helm/community/blob/main/hips/hip-0006.md) compliant HELM Charts in the Downloader when interpreting HELM Chart Artifact Types.

This will especially fix resource downloads which previously assumed layer 0, which is incorrect. Instead we have to find the first and only occurrence of `application/vnd.cncf.helm.chart.content.v1.tar+gzip`. Similarly, finding provenance data should not be hardcoded to layer 1, but to the first and only occurrence of `application/vnd.cncf.helm.chart.provenance.v1.prov`

Fix https://github.com/open-component-model/ocm/issues/1469

### DIFF
--- a/api/ocm/extensions/download/handlers/helm/handler.go
+++ b/api/ocm/extensions/download/handlers/helm/handler.go
@@ -116,7 +116,7 @@ func (h Handler) Download(p common.Printer, racc cpi.ResourceAccess, path string
 // download downloads the chart and optional provenance file from an  oci.ArtifactAccess.
 // the format of the artifact is expected to match the official HELM Reference Specification
 //
-// see https://github.com/helm/community/blob/main/hips/hip-0006.md
+// see https://github.com/helm/community/blob/dd5fe7878e293c573cc22db5d36558709c7b8a43/hips/hip-0006.md
 func download(p common.Printer, art oci.ArtifactAccess, path string, fs vfs.FileSystem) (chart, prov string, err error) {
 	var finalize finalizer.Finalizer
 	defer finalize.FinalizeWithErrorPropagation(&err)


api/ocm/extensions/download/handlers/helm/handler.go
```diff
@@ -101,41 +113,72 @@ func (h Handler) Download(p common.Printer, racc cpi.ResourceAccess, path string
 	return h.fromOCIArtifact(p, meth, path, fs)
 }
 
+// download downloads the chart and optional provenance file from an  oci.ArtifactAccess.
+// the format of the artifact is expected to match the official HELM Reference Specification
+//
+// see https://github.com/helm/community/blob/main/hips/hip-0006.md
```
frewilhelm, line 119: I am more of a fan of perma-links, but I guess that is debatable
```suggestion
// see https://github.com/helm/community/blob/dd5fe7878e293c573cc22db5d36558709c7b8a43/hips/hip-0006.md
```

frewilhelm (approved):
lgtm in general

Although, binary data (== .tgz file) should not be part of the repository. However, as we already have binary data in this repository, I am fine with that for now.
But I'd like to discuss at some point how we can avoid this in the future and other repositories

jakobmoellerdev:
I would think that sometimes tests must work with preset testdata because the order that we have in this test case is literally unachievable within the current lib without a custom script.

Skarlso (approved):

Issue #1474, opened 2025-05-20
GitHub actions workflow `Markdown` is failing


frewilhelm:
**Context**  
Currently, the GitHub actions `Markdown` is [failing](https://github.com/open-component-model/ocm/actions/workflows/verify-md.yml) due to a malformatted regex from the config-file.

Strangely enough, the workflow is only failing since some weeks while the regex is present since six month.

However, this is all irrelevant because the original GitHub action got a deprecation note since April 2025. Accordingly, the workflow must be adjusted either way.

TODOs:
- [ ] Adjust the workflow to verify markdown-links again

PR #1475, opened 2025-05-21
fix: use the updated link checker fork

Skarlso:

### DIFF
--- a/.github/workflows/verify-md.yml
+++ b/.github/workflows/verify-md.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       - name: Markdown verify links
-        uses: gaurav-nelson/github-action-markdown-link-check@3c3b66f1f7d0900e37b71eca45b63ea9eedfce31 # 1.0.17
+        uses: tcort/github-action-markdown-link-check@a800ad5f1c35bf61987946fd31c15726a1c9f2ba # v1.1.0
         with:
           use-quiet-mode: yes
           use-verbose-mode: no


frewilhelm:
This will fix https://github.com/open-component-model/ocm/issues/1474

frewilhelm:
Did you had any chance to check if this works?

frewilhelm:
Yes, I was wondering about the same, when I took a look yesterday. Apparently, it was working fine until the update of the original action.

Skarlso:
Okay, the check is now fixed. Funny enough `,{ "pattern": "^{{.*}}$" }` we had in there actually doesn't work, because it's being translated into the regex I added now.

.github/config/markdown-link-check.json
```diff
@@ -5,6 +5,7 @@
      { "pattern": "^http[s]*://localhost.*" }
     ,{ "pattern": "^https://.*\\.sigstore\\.dev.*" }
     ,{ "pattern": "^{{.*}}$" }
-    ,{ "pattern": "^https://github.com/**/compare/**" }
+    ,{ "pattern": "^https://github.com/.*/compare/.*" }
+    ,{ "pattern": "%7B%7B.*%7D%7D" }
```
Skarlso, line 9: let it be known now and for future reference.

The above `"^{{.*}}$"` is actually not working, because the link being checked is URL encoded BEFORE checking. Meaning the actual pattern of the link that we'd need to ignore in case of our lovely templated links is this one.

Skarlso:
Previous failed run: https://github.com/open-component-model/ocm/actions/runs/15159082663/job/42620755562
Previous passed run when I edited a markdown: https://github.com/open-component-model/ocm/actions/runs/15159137065

jakobmoellerdev (approved):

PR #1501, opened 2025-06-06
fix!: unwrap standard transfer handler and fix spiff handler

### LINKED ISSUE #1500: Different transfer behaviour when using `ocm transfer ctf/component/?` (with options)
**Context**  
Initially, we noticed that `ocm transfer component` ignores the `--omit-access-type` flag. We did figure out that this is due to a missing check in the `spiff` transfer handler:

[`spiff`](https://github.com/open-component-model/ocm/blob/c2006477ffd078787d488b4bdea870a015df5d4b/api/ocm/tools/transfer/transferhandler/spiff/handler.go#L111C1-L120C2):
```go
func (h *Handler) TransferResource(src ocm.ComponentVersionAccess, a ocm.AccessSpec, r ocm.ResourceAccess) (bool, error) {
	if !h.opts.IsResourcesByValue() {
		return false, nil
	}
	if h.opts.GetScript() == nil {
		return true, nil
	}
	binding := h.getBinding(src, a, &r.Meta().ElementMeta, &r.Meta().Type, nil)
	return h.EvalBool("resource", binding, "process")
}
```

in comparison to the [`standard`](https://github.com/open-component-model/ocm/blob/c2006477ffd078787d488b4bdea870a015df5d4b/api/ocm/tools/transfer/transferhandler/standard/handler.go#L64) handler:
```go
func (h *Handler) TransferResource(src ocm.ComponentVersionAccess, a ocm.AccessSpec, r ocm.ResourceAccess) (bool, error) {
	if h.opts.IsAccessTypeOmitted(a.GetType()) {
		return false, nil
	}
	if h.opts.IsLocalResourcesByValue() {
		if r.Meta().Relation == metav1.LocalRelation {
			return true, nil
		}
	}
	return h.opts.IsResourcesByValue(), nil
}
```

This check is missing:
```go
	if h.opts.IsAccessTypeOmitted(a.GetType()) {
		return false, nil
	}
```

While debugging this, however, we noticed that this `ocm transfer ctf` acknowledges the `--omit-access-type` flag because it defaults to the `standard` transfer-handler.

Accordingly, there is one bug in the `spiff`-handler wrt to the check but also depending on the command, a different transfer handler is chosen in general. So far, we only checked this for `ctf` and `component`. Maybe this is also true for other objects.

**Expected behavior**  
A consistent and understandable call of the transfer handlers as well as respecting any passed option.

(This bug is a result of a joined discussion and is used as tracker. @discussion-attendees feel free to edit this description if necessary)

fabianburth:

- Make `ocm transfer {cv|ctf}` respect `--omit-access-types` (as originally intended).
- Unwrap the standard transfer handler so omit logic is applied consistently across handlers.
- Fix spiff handler so it no longer reinstates access types when omit is requested. 

Previously, users might have used the flag `--omit-access-types` with a transfer and would have changed behavior now (flag was ignored). 

After this change, `--omit-access-types` takes effect for any transfer as originally intended. 

**Users expecting resources to be transferred even though they were mentioned with their access type in  `--omit-access-types` must likely drop `--omit-access-types` for that access type if they used it before and expected the resources to still be present.**

Fixes #1500

### DIFF
--- a/api/ocm/tools/transfer/transferhandler/spiff/handler.go
+++ b/api/ocm/tools/transfer/transferhandler/spiff/handler.go
@@ -112,6 +112,9 @@ func (h *Handler) TransferResource(src ocm.ComponentVersionAccess, a ocm.AccessS
 	if !h.opts.IsResourcesByValue() {
 		return false, nil
 	}
+	if h.opts.IsAccessTypeOmitted(a.GetType()) {
+		return false, nil
+	}
 	if h.opts.GetScript() == nil {
 		return true, nil
 	}
@@ -121,6 +124,9 @@ func (h *Handler) TransferResource(src ocm.ComponentVersionAccess, a ocm.AccessS
 
 func (h *Handler) TransferSource(src ocm.ComponentVersionAccess, a ocm.AccessSpec, r ocm.SourceAccess) (bool, error) {
 	if !h.opts.IsSourcesByValue() {
+		return false, nil
+	}
+	if h.opts.IsAccessTypeOmitted(a.GetType()) {
 		return false, nil
 	}
 	if h.opts.GetScript() == nil {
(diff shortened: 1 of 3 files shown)


fabianburth:
well, that's unlucky

PR #1548, opened 2025-07-20
fix: double slash in `imageReference`

ikhandamirov:

This PR fixes the problem described below.

Steps to reproduce.
1. Prepare the following for the `ocm transfer cv` command:
    1.1. An OCM component having a resource of type `ociImage`
    1.2. An OCI registry to transfer the component to
2. Use the following parameters for the transfer operation: `--copy-resources --uploader 'ocm/ociArtifacts={"ociRef":"<TARGET_OCI_REGISTRY>//"}'`

Note the problem:
1. The parameters instruct OCM CLI to use a named upload handler to copy OCI artifacts to the root of the specified registry.
2. View OCM component descriptor of the copied component in the target registry and note that the image reference to the OCI resource has `//` between registry host and artifact namespace. The reference is invalid, as there should be only one `/`.


jakobmoellerdev (commented):
@fabianburth and I looked at this again and we might want to revisit this in the following way:

basically we checked the OCI generic handler tests and noticed various TODOs from uwe that look like this:

`// TODO: the result is invalid for ctf: better handling for ctf refs`

this is because the generic handler uses invalid mappings and does "best-effort" conversion when working with OCM repo specs.

we can enforce a stricter mapping to OCI repositories based on `oci.ParseRef` for cases in which we have everything available (adjustment in https://github.com/open-component-model/ocm/blob/69ea53a86ed3ae39e815f9bc73014bf49f3c731e/api/ocm/extensions/blobhandler/handlers/generic/ocirepo/blobhandler.go):

```go
	suffix := tag
	if version != "" {
		suffix += version
	}

	host, port := base.HostPort()
	repoRef := host
	if port != "" {
		repoRef += ":" + port
	}
	if namespace.GetNamespace() != "" {
		repoRef += "/" + namespace.GetNamespace()
	}
	artRaw := repoRef + suffix

	artRef, err := oci.ParseArt(artRaw) // validate the reference
	if err != nil {
		// if this not an official OCI reference, we still allow it (for legacy reasons)
		// but return it as an OCM OCI Artifact Reference (note that this is technically not an OCI reference and invalid)
		return ociartifact.New(base.ComposeRef(namespace.GetNamespace() + tag + version)), nil
	}

	return ociartifact.New(artRef.String()), nil
```

This makes it so that legacy behavior still works like before, but if we encounter OCI references, we will try to parse them as OCI Artifact References first. 

That way we can also avoid adding a `//` parsing as optional for the artifact regexp which we should avoid touching at all costs to avoid regressions in other parts of the codebase.

jakobmoellerdev:
The fix is above. You were not part of the discussion because it was adhoc, feel free to open up new sessions if you need further feedback. 

The reason im hesitant on the regex fix is because you are opening a Regex for something that is not meant to support the scenario you are opening. Even in OCM a oci reference such as `ghcr.io/user/prefix//` is not something we would recommend.

ikhandamirov:
@jakobmoellerdev, your version works for me. Is in.

fabianburth:
@ikhandamirov while I participated in the fix proposed by @jakobmoellerdev and am confident it should work, I think we should still add some tests

ikhandamirov:
@fabianburth , I'm confident that the fix solves the particular issue, because I tested it manually. 

Regarding automated tests, the question is the following. The test scenario involves `ocm tansfer` to an OCI registry. Do we already have such tests in OCM V1? Or would it be the first one? Is there a test component and a test registry that can be reused?

The fix is in `api/ocm/extensions/blobhandler/handlers/generic/ocirepo/blobhandler.go`, the respective test suite is `api/ocm/extensions/blobhandler/handlers/generic/ocirepo/upload_test.go`, but it works with CTFs.

jakobmoellerdev:
👍  this is desired state as ociRef is probably not the best format for the use case we are targeting anyways.

api/ocm/extensions/blobhandler/handlers/generic/ocirepo/blobhandler.go
```diff
@@ -135,27 +135,44 @@ func (b *artifactHandler) StoreBlob(blob cpi.BlobAccess, artType, hint string, g
...
+	if namespace != "" {
+		repoRef += "/" + namespace
+	} else {
+		// TODO(ikhandamirov): remove this workaround once the oci.ParseArt() function is fixed.
+		return "", errors.ErrInvalid("namespace cannot be empty in an OCI reference")
 	}
 	artRaw := repoRef + suffix
 
+	// TODO(ikhandamirov): oci.ParseArt() as a validating function has certain shortcomings.
+	// 1. If a port number is specified, it returns an error, even though the port is allowed in an OCI reference.
```
jakobmoellerdev, line 168: instead of using oci.ParseArt you may want to use the oras tooling here for proper ref parsing

fabianburth (changes requested):
Comment above

jakobmoellerdev (commented):
Mostly LGTM

fabianburth (dismissed):

jakobmoellerdev (approved):

PR #1588, opened 2025-08-22
chore: update Go version to 1.25 and update some modules

hilmarf:
except: k8s.io/kube-openapi
It introduces a dependency on sigs.k8s.io/structured-merge-diff/v6, which causes issues.

follow up on: https://github.com/open-component-model/ocm/pull/1586

Fixes https://github.com/open-component-model/ocm-project/issues/604


jakobmoellerdev:
please bump the minimum required go version and the dependencies individually. then we can revert any problematic patches atomically

jakobmoellerdev:
BTW The test failure is from your bump to mandelsoft/go-utils and mandelsoft/vfs. Please revert those, then the tests will succeed again :)

hilmarf:
https://github.com/open-component-model/ocm/pull/1591

PR #1596, opened 2025-08-28
fix: go 1.25 only available in nix-unstable

vasu1124:
go 1.25 only available in nix-unstable


jakobmoellerdev (commented):
I am aware of nix builds failing but I don't think we want to publish a nix build from unstable. Thus, we should rather wait here until nix publishes a version with go1.25. WDYT?

PR #1609, opened 2025-09-09
chore: update Go version to 1.25.1

fabianburth:

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,8 +144,7 @@ You can pass in the following arguments to override the predefined defaults:
 Please check [hub.docker.com](https://hub.docker.com/_/golang/tags?page=1&name=alpine) for possible version combinations.
 
 ```bash
-podman build -t ocm --build-arg GO_VERSION=1.25.1 --build-arg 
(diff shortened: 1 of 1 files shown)


components/helminstaller/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25.1 AS builder
```
hilmarf, line 1: I would not pin the container version... as it normally is updated automatically to the latest available patch version 

Image

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM golang:1.25-alpine3.22 AS build
+FROM --platform=$BUILDPLATFORM golang:1.25.1-alpine3.22 AS build
```
hilmarf, line 1: I would not pin the container version... as it normally is updated automatically to the latest available patch version

PR #1646, opened 2025-10-14
feat: support index based reading of OCI artifacts

jakobmoellerdev:

Refactored `ComponentVersionContainer` to allow fetching OCI manifests from top level OCI Image indexes. This is preparation for allowing the reference of native OCI Manifests that are themselves referenced via index first. This is prep work to allow native storage in OCI registries for local blobs that might themselves be stored previously as OCM Artifact Sets. They could now be accessed natively instead.

This allows native reading of OCM Artifacts created with the new v2 library that can use indexes to store artifacts natively

final stepping stone to prepare ADR for https://github.com/open-component-model/ocm-project/issues/680
fixes https://github.com/open-component-model/ocm-project/issues/717

### DIFF
--- a/api/oci/extensions/repositories/artifactset/ctf_roundtrip_test.go
+++ b/api/oci/extensions/repositories/artifactset/ctf_roundtrip_test.go
@@ -28,7 +28,7 @@ const (
 	componentVersion = "1.0.0"
 	resourceName     = "hello-image"
 	resourceVersion  = "1.0.0"
-	imageReference   = "ghcr.io/piotrjanik/open-component-model/hello-ocm:latest"
+	imageReference   = "ghcr.io/open-component-model/open-component-model/hello-ocm:latest"
 )
 
 func gunzipToTar(tgzPath string) (string, error) {


api/oci/extensions/repositories/artifactset/artifactset.go
```diff
@@ -204,26 +206,123 @@ func (a *namespaceContainer) AddTags(digest digest.Digest, tags ...string) error
...
+		}
+	}
+	if len(descIdxs) == 0 {
+		return errors.ErrUnknown(cpi.KIND_OCIARTIFACT, digest.String())
+	}
+
+	// Build complete tag set from:
+	//   - existing TAGS_ANNOTATION on any descriptor for this digest
+	//   - incoming tags
+	tagSet := map[string]struct{}{}
```
frewilhelm: From what I can see this could also be just a `[]string`. Am I missing something?

api/oci/extensions/repositories/artifactset/artifactset.go
```diff
@@ -204,26 +206,123 @@ func (a *namespaceContainer) AddTags(digest digest.Digest, tags ...string) error
...
-			} else {
-				cur = strings.Join(tags, ",")
-			}
-			e.Annotations[TAGS_ANNOTATION] = cur
-			if a.base.FileSystemBlobAccess.Access().GetInfo().GetDescriptorFileName() == OCIArtifactSetDescriptorFileName {
-				e.Annotations[OCITAG_ANNOTATION] = tags[0]
+	isOCI := a.base.FileSystemBlobAccess.
+		Access().
+		GetInfo().
+		GetDescriptorFileName() == OCIArtifactSetDescriptorFileName
```
frewilhelm: Just structure-wise - this variable could be set right before it is used on L256

api/ocm/extensions/repositories/genericocireg/accessmethod_localblob.go
```diff
@@ -40,6 +43,11 @@ func newLocalBlobAccessMethodImpl(a *localblob.AccessSpec, ns oci.NamespaceAcces
 		namespace: ns,
 		artifact:  art,
 	}
+	if m.spec.MediaType == artdesc.MediaTypeImageIndex || m.spec.MediaType == artdesc.MediaTypeImageManifest {
+		// if we discover a localblob with an index or manifest media type, we can
+		// assume that we are dealing with a new style of artifact created by the new reference library.
+		m.mimeType = artifactset.MediaType(m.spec.MediaType)
+	}
```
frewilhelm: What if the localblob would be an actual OCI artifact? This would not make a difference as the new reference library creates its artifact in the OCI layout, right?

api/oci/extensions/repositories/artifactset/artifactset.go
jakobmoellerdev: the point of this was that i am able to unify both tags from the vararg and the annotation without having to deduplicate

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -89,11 +132,20 @@ func (c *ComponentVersionContainer) GetParentBridge() repocpi.ComponentAccessBri
...
+
+	var err error
+	if c.indexArtifact != nil {
+		err = errors.Join(err, c.indexArtifact.Close())
+	}
+	if c.manifestArtifact != nil {
+		err = errors.Join(err, c.manifestArtifact.Close())
+	}
+	return err
 }
```
Skarlso: This is now not calling `c.access.Close()`. That's intentional, right?

api/oci/extensions/repositories/ctf/namespace.go
```diff
@@ -40,10 +40,6 @@ func (n *namespaceContainer) Close() error {
 	return nil
 }
 
-func (n *namespaceContainer) GetBlobDescriptor(digest digest.Digest) *cpi.Descriptor {
```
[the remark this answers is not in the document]
Skarlso: Okay then :D

api/oci/extensions/repositories/artifactset/utils_synthesis.go
```diff
@@ -104,9 +110,8 @@ func SynthesizeArtifactBlobForArtifact(art cpi.ArtifactAccess, ref string, filte
 			return "", fmt.Errorf("failed to transfer artifact: %w", err)
 		}
 
-		if ok := vers.IsTagged(); ok {
-			err = set.AddTags(*dig, vers.GetTag())
-			if err != nil {
+		if len(tags) > 0 {
```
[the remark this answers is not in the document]
jakobmoellerdev: how does that help me here? what do you want me to do?

api/ocm/extensions/repositories/genericocireg/componentversion.go
```diff
@@ -46,38 +46,81 @@ func newComponentVersionAccess(mode accessobj.AccessMode, comp *componentAccessI
...
+		m = access.ManifestAccess()
+		if m == nil {
+			return nil, fmt.Errorf("artifact is neither manifest nor index")
+		}
+
+		manifestArtifact = access
 	}
-	state, err := NewState(mode, comp.name, version, m, compatattr.Get(comp.GetContext()))
+
+	state, err := NewState(mode, comp.name, version, i, m, compatattr.Get(comp.GetContext()))
```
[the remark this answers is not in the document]
jakobmoellerdev: good catch!

PR #1676, opened 2025-11-12
feat(transfer): Optimized approach for OCM transfer by implementing a concurrent worker pool

jakobmoellerdev:

Introduces a new experimental transfer mode that can be set with a max worker attribute. Default is still sequential. Concurrent transfer is enabled as soon as workers > 1

fix https://github.com/open-component-model/ocm-project/issues/740


Skarlso (commented):
Some questions. :)

jakobmoellerdev:
I believe the latest change made the concurrent transfer race on file systems, still debugging

PR #1726, opened 2025-12-05
fix!: use Fulcio certificate instead of public key and upgrade Sigstore Cosign from v2 to v3

morri-son:
On-behalf-of: Gerald Morrison (SAP) <gerald.morrison@sap.com>

This PR introduces two breaking changes:

## Breaking change 1: Correct Sigstore bundle content in signatures
OCM now bundles the short‑lived Fulcio certificate instead of just a raw public key into signatures when using keyless signing. To preserve backwards compatibility for existing components, we introduced a new signing algorithm `sigstore-v2` and retain the legacy `sigstore` algorithm:

- `sigstore`: Stores only the public key in the Rekor entry. Existing signatures remain verifiable.
- `sigstore-v2`: Stores the Fulcio certificate in the Rekor entry, conforming to the Sigstore Bundle spec. Recommended algorithm to be used for new signatures.

### OCM CLI Examples
Create a new signature using `sigstore-v2`:
`ocm sign componentversion --signature mysig --algorithm sigstore-v2 --keyless <component-version>`

Verify a signature (correct algorithm is determined from signature name `mysig`)
`ocm verify componentversion --signature mysig --keyless <component-version>`

## Breaking change 2: Cosign v3 changes OIDC token handling in GitHub Actions

GitHub no longer auto‑injects an ID token when `SIGSTORE_ID_TOKEN` is missing; workflows must explicitly request and export it.

Typical CI error if `SIGSTORE_ID_TOKEN` is missing:
`getting ID token: executing OIDC flow: failed to start browser`

### Migration on GitHub Actions

Add `id-token: write` permission and explicitly request an OIDC ID token with audience `sigstore`, then export it as `SIGSTORE_ID_TOKEN` prior to any `ocm sign componentversion --keyless` steps.

Below you find a snippet from an example workflow step that gets an OIDC token and exports it to the GitHub env for further use.

```yaml
- name: Acquire GitHub OIDC ID token
  shell: bash
  run: |
    TOKEN=$(curl -sSf \
      -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
      "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=sigstore" \
      | jq -r '.value')

    echo "::add-mask::$TOKEN"
    echo "SIGSTORE_ID_TOKEN=$TOKEN" >> "$GITHUB_ENV"
```

### Other CI platforms are not affected

This problem is specific to GitHub Actions. Other CI/CD platforms have always required explicit handling of OIDC tokens and environment variables for Sigstore keyless flows. Therefore, no migration is needed outside GitHub Actions.

## Recommended path forward

- Continue to verify existing `sigstore` signatures — they remain valid.
- For new signatures, use `sigstore-v2` to ensure Sigstore Bundle compliance.
- On GitHub Actions, add explicit OIDC token handling and export `SIGSTORE_ID_TOKEN` as shown above.

## References

- Sigstore OIDC Documentation: https://docs.sigstore.dev/cosign/oidc/
- Cosign v3 Release Notes: https://github.com/sigstore/cosign/releases/tag/v3.0.0

### DIFF
--- a/api/tech/signing/handlers/sigstore/handler.go
+++ b/api/tech/signing/handlers/sigstore/handler.go
@@ -320,30 +320,27 @@ func extractECDSAPublicKey(pubKeyBytes []byte) (*ecdsa.PublicKey, error) {
 	if block == nil {
 		return nil, fmt.Errorf("no PEM block found in Fulcio public key")
 	}
+	var rawPub any
 	switch block.Type {
 	case "PUBLIC KEY":
 		result, err := x509.ParsePKIXPublicKey(block.Bytes)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse public key: %w", err)
 		}
-		// cast to ecdsa.PublicKey as we use this in the verification
-		pub, ok := result.(*ecdsa.PublicKey)
-		if !ok {
-			return nil, fmt.Errorf("unexpected public key type: %T", result)
-		}
-		return pub, nil
+		rawPub = result
 	case "CERTIFICATE":
 		cert, err := x509.ParseCertificate(block.Bytes)
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse Fulcio certificate: %w", err)
 		}
-		// cast to ecdsa.PublicKey as we use this in the verification
-		pub, ok := cert.PublicKey.(*ecdsa.PublicKey)
-		if !ok {
-			return nil, fmt.Errorf("unexpected certificate public key type: %T", cert.PublicKey)
-		}
-		return pub, nil
+		rawPub = cert.PublicKey
 	default:
 		return nil, fmt.Errorf("unsupported PEM block type: %s", block.Type)
 	}
+	// cast to ecdsa.PublicKey as we use this in the verification
+	pubKey, ok := rawPub.(*ecdsa.PublicKey)
+	if !ok {
+		return nil, fmt.Errorf("unexpected public key type: %T", rawPub)
+	}
+	return pubKey, nil
 }


api/tech/signing/handlers/sigstore/handler.go
```diff
@@ -31,28 +31,37 @@ import (
 )
 
 // Algorithm defines the type for the RSA PKCS #1 v1.5 signature algorithm.
-const Algorithm = "sigstore"
+// Since "sigstore" contained a buggy implementation, we are introducing "sigstore-v3" as well.
+// "sigstore-v3" correctly injects the Fulcio fs.Cert into the bundle, not just the public key like in "sigstore".
+const (
+	Algorithm   = "sigstore"
+	AlgorithmV3 = "sigstore-v3"
```
jakobmoellerdev: this algo name is misleading. v3 is the name of the lib, but this is independent of the bump. the algo should reflect your actual change and would warrant something like `sigstore-cert` or `sigstore-v2`

api/tech/signing/handlers/sigstore/handler.go
```diff
@@ -295,3 +315,36 @@ func prepareRekorEntry(digest string, sig, publicKey []byte) hashedrekord_v001.V
 		},
 	}
 }
+
+func extractECDSAPublicKey(pubKeyBytes []byte) (*ecdsa.PublicKey, error) {
+	block, _ := pem.Decode(pubKeyBytes)
```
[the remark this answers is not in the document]
morri-son: The public key material that we get from rekor has been created by ourselves during signing. For the "sigstore" case we create exactly one PEM block ourselves using pem.EncodeToMemory().

For "sigstore-v2" we use the fs.Cert to create it and the fs.Cert only contains the Fulcio leaf certificate, which is exactly one cert in PEM format. fs.Chain contains the whole cert chain, but it is not used by us. So, we should never hit the edge case you described.

api/tech/signing/handlers/sigstore/handler.go
```diff
@@ -137,8 +152,17 @@ func (h Handler) Sign(cctx credentials.Context, digest string, sctx signing.Sign
 		return nil, fmt.Errorf("failed to create rekor client: %w", err)
 	}
 
+	// decide which public material to use for rekor entry
+	// old "sigstore" flow uses only public key
+	// new "sigstore-v2" flow uses the fulcio certificate
+	// Since v3 the Fulcio certificate fs.Cert is already in PEM format
+	rekorPublicMaterial := publicKey
+	if h.Algorithm() == AlgorithmV2 {
```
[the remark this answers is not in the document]
morri-son: 👍🏻 I moved all actions for "sigstore" pub key handling into the if clause. Now they are not executed when wir on algo V2.

morri-son:
all comments answered and feedback integrated.

api/tech/signing/handlers/sigstore/handler.go
```diff
@@ -295,3 +313,38 @@ func prepareRekorEntry(digest string, sig, publicKey []byte) hashedrekord_v001.V
...
+		}
+		// cast to ecdsa.PublicKey as we use this in the verification
+		pub, ok := cert.PublicKey.(*ecdsa.PublicKey)
+		if !ok {
+			return nil, fmt.Errorf("unexpected certificate public key type: %T", cert.PublicKey)
+		}
+		return pub, nil
+	default:
+		return nil, fmt.Errorf("unsupported PEM block type: %s", block.Type)
+	}
```
fabianburth: could be simplified / shortened by deduplicating the type assertion below the switch case

api/tech/signing/handlers/sigstore/handler_test.go
```diff
@@ -0,0 +1,165 @@
...
+	_, err := extractECDSAPublicKey(unsupportedPEM)
+
+	assert.Error(t, err)
+}
+
+// ================================================================
+// Verify signatures with both Sigstore algorithms (works offline,
+// as Rekor public keys are embedded in Cosign library and
+// all verification data contained in Sigstore bundle)
+// ================================================================
```
fabianburth: maybe remove the =================================

api/tech/signing/handlers/sigstore/testdata/component-descriptor-signed.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - digest:
+      hashAlgorithm: SHA-256
+      normalisationAlgorithm: jsonNormalisation/v3
+      value: 20d12fdfaf1f57c970d63608db4eaeaab934527ab19caac9b4518502d31c5d72
+    name: sigstore-recommended
+    signature:
+      algorithm: sigstore-v3
```
fabianburth: is this correct?

api/tech/signing/handlers/sigstore/handler_test.go
```diff
@@ -0,0 +1,165 @@
...
+// Test error handling for unsupported PEM type
+func TestExtractECDSAPublicKey_UnsupportedPEMType(t *testing.T) {
+	unsupportedPEM := pem.EncodeToMemory(&pem.Block{
+		Type:  "UNSUPPORTED",
+		Bytes: []byte("some data"),
+	})
+
+	_, err := extractECDSAPublicKey(unsupportedPEM)
+
+	assert.Error(t, err)
```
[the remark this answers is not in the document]
morri-son: replaced assert.Error with assert.ErrorContains or assert.EqualError

api/tech/signing/handlers/sigstore/handler_test.go
```diff
@@ -0,0 +1,165 @@
...
+// ================================================================
+
+// Verify legacy "sigstore" signature
+func TestVerify_LegacySignature(t *testing.T) {
+	descriptorYAML := loadTestData(t, "component-descriptor-signed.yaml")
+	digest, sigValue := getSignatureByName(t, descriptorYAML, "sigstore-legacy")
+
+	handler := Handler{algorithm: Algorithm}
+
+	err := handler.Verify(digest, &signing.Signature{
```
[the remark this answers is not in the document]
morri-son: added two checks, one for digest mismatch and one for an invalid signature.

api/tech/signing/handlers/sigstore/handler.go
morri-son: Simplfied like proposed.

PR #1815, opened 2026-02-09
chore: correct brew and readme 

morri-son:
On-behalf-of: Gerald Morrison (SAP) <gerald.morrison@sap.com>

correct brew and readme


frewilhelm (changes requested):
the PR title differs from what is actually changed

morri-son:
> the PR title differs from what is actually changed

already changed

Issue #1731, opened 2025-12-10
Add timeout configuration flag to OCM CLI


imko:
**Description**
**What would you like to be added?** 
We would like to request adding a configurable timeout flag to the OCM CLI.

When we run commands like `ocm get cv ...` ([ref](https://github.com/open-component-model/ocm/blob/main/docs/reference/ocm_get_componentversions.md)) from China, the CLI often times out while accessing component versions from global repositories because of slow and unstable network conditions. We tried adding our own retry logic and a timeout in our code, but the CLI’s internal timeout triggers first, so the retries never get a chance to run successfully. 

Adding a timeout flag (e.g., --timeout) would allow users to increase or customize the timeout duration based on their network environment. This would make the CLI more reliable in regions with limited and weaker connectivity and avoid premature failures. 

**Why is this needed?**
* The built-in timeout can’t be adjusted and fires before external retry logic can handle it.
* Network conditions in some regions require a longer timeout than the default.
* A configurable timeout would significantly improve usability and robustness for global users.

Thanks for considering this feature request.

piotrjanik:
implemented new config type: http.config.ocm.software/v1alpha1
https://github.com/open-component-model/open-component-model/pull/1770

piotrjanik:
PR for v1 got accepted.
PR for v2 in progress

imko:
Hi @piotrjanik and @frewilhelm,

Just following up to see if this issue has been added back to your sprint and if there's an updated ETA. Thanks!

PR #1823, opened 2026-02-13
feat: add registry client timeout attribute

piotrjanik:

Adds a configurable timeout attribute preventing the CLI from terminating connections too early when interacting with slower OCI or Docker registries.

- Introduces a new `http.config.ocm.software/v1alpha1` config type and `timeout` attribute for configuring client timeouts (default: 30s).
- Adds global `--timeout` flag to the OCM CLI that applies the timeout via config context.

**Integration test**:
- Uses `toxiproxy` + `registry:2.8.3` testcontainers on a shared Docker network to simulate latency
- Verifies transfer fails with a timeout shorter than proxy latency (2s timeout, 30s latency)
- Verifies transfer succeeds with a timeout exceeding proxy latency (30s timeout, 1s latency)

Fixes: https://github.com/open-component-model/ocm/issues/1731


matthiasbruns (dismissed):
love it - toxiproxy for the win :D

Skarlso (approved):

PR #1887, opened 2026-03-30
feat: add configurable HTTP client timeouts via config file

### LINKED ISSUE #1731: Add timeout configuration flag to OCM CLI

piotrjanik:

Adds a new `http.config.ocm.software` config type that allows users to configure HTTP client timeout settings for OCI registry and Docker daemon connections. This prevents requests from hanging indefinitely on slow or broken networks.

Supported timeout fields (all optional, Go duration strings like "30s", "5m"):
- timeout — overall http.Client deadline
- tcpDialTimeout — TCP connection establishment
- tcpKeepAlive — TCP keep-alive probe interval
- tlsHandshakeTimeout — TLS handshake
- responseHeaderTimeout — waiting for response headers
- idleConnTimeout — idle connection lifetime

Design:
- Follows the regular OCI config pattern (like credentials config) — HTTPSettings is stored on the OCI context via GetHTTPSettings()/SetHTTPSettings(), not as a datacontext attribute.
- Duration is a validated string type with TimeDuration() returning (*time.Duration, error) — nil means "not configured" (use default), distinct from zero ("explicitly disabled").
- httpclient.NewTransport() clones http.DefaultTransport and selectively overrides only the configured timeouts.
- Integration tests use Toxiproxy to verify timeout behavior against a real registry with injected latency.

Fixes: https://github.com/open-component-model/ocm/issues/1731


jakobmoellerdev (dismissed):

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -153,12 +154,19 @@ func (r *RepositoryImpl) getResolver(comp string) (oras.Resolver, error) {
...
+	timeout, err := httpCfg.GetTimeout()
+	if err != nil {
+		return nil, err
+	}
+
 	if r.info.Scheme == "https" {
-		// set up TLS
 		//nolint:gosec // used like the default, there are OCI servers (quay.io) not working with min version.
-		conf := &tls.Config{
+		baseTransport.TLSClientConfig = &tls.Config{
```
Skarlso: Errr... aren't you overwriting any user settings here that would exist and be set coming from `NewTransport`?

Skarlso (commented):
Largely okay, just spotted a few things here :)

api/oci/internal/context.go
```diff
@@ -193,3 +197,36 @@ func (c *_context) SetAlias(name string, spec RepositorySpec) {
 	defer c.updater.Unlock()
 	c.aliases[name] = spec
 }
+
+func (c *_context) GetHTTPSettings() *HTTPSettings {
```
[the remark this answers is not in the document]
piotrjanik: fixed

api/oci/internal/httpsettings.go
```diff
@@ -0,0 +1,124 @@
...
+func (s *HTTPSettings) GetTCPDialTimeout() (*time.Duration, error) {
+	if s == nil {
+		return nil, nil
+	}
+	return s.TCPDialTimeout.TimeDuration()
+}
+
+// GetTCPKeepAlive returns the TCP keep-alive interval.
+// Returns (nil, nil) if not set.
+func (s *HTTPSettings) GetTCPKeepAlive() (*time.Duration, error) {
```
jakobmoellerdev: these getters are imho not required. especially the nil check

api/oci/internal/httpsettings.go
```diff
@@ -0,0 +1,119 @@
...
+
+// nonNegative requires the duration to be zero or positive.
+func nonNegative(d Duration) bool {
+	return !strings.HasPrefix(string(d), "-")
+}
+
+// nonNegativeOrMinusOne requires the duration to be zero, positive,
+// or -1 (to disable keep-alive probes).
+func nonNegativeOrMinusOne(d Duration) bool {
+	return !strings.HasPrefix(string(d), "-") || strings.HasPrefix(string(d), "-1")
```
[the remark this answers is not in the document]
piotrjanik: No, it does not.
I changed validation:
1. Strings like `1m` are parsed and converted to Duration - if that does not work, error is returned. Weird values like "1g5h" would be rejected here.
2. It returns an error when the value is negative (tcp keep alive is an exception, as any negative value turns it off)

api/oci/extensions/repositories/ocireg/timeout_integration_test.go
```diff
@@ -0,0 +1,230 @@
...
+		Expect(err.Error()).To(ContainSubstring("i/o timeout"))
+	})
+
+})
+
+func addLatency(proxy *toxiproxy.Proxy, latencyMs int, stream string) {
+	_, err := proxy.AddToxic("latency", "latency", stream, 1.0, toxiproxy.Attributes{
+		"latency": latencyMs,
+	})
+	Expect(err).To(Succeed())
```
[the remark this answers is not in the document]
piotrjanik: done

api/oci/config/httpconfig.go
```diff
@@ -0,0 +1,109 @@
...
+type HTTPConfig struct {
+	runtime.ObjectVersionedType `json:",inline"`
+	cpi.HTTPSettings            `json:",inline"`
+}
+
+func (a *HTTPConfig) GetType() string {
+	return HTTPConfigType
+}
+
+func (a *HTTPConfig) ApplyTo(_ cfgcpi.Context, target interface{}) error {
```
jakobmoellerdev, line 29: why do we need this strange interface?

api/oci/extensions/repositories/ocireg/repository.go
```diff
@@ -197,6 +189,40 @@ func (r *RepositoryImpl) getResolver(comp string) (oras.Resolver, error) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	timeout, err := httpCfg.Timeout.TimeDuration()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if scheme == "https" {
+		//nolint:gosec // there are OCI servers (quay.io) not working with min version.
```
jakobmoellerdev: shouldnt the min version be configurable as well?

api/oci/internal/httpsettings.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// Duration is a string type representing a Go duration (e.g. "30s", "5m").
+// It is validated on JSON unmarshaling.
+type Duration string
```
jakobmoellerdev: i know we went over this a few times, but why is Duration its own time aliased to `string`? I was expecting `type Duration time.Duration` here.

api/oci/internal/httpsettings.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, nil
+	}
+	pd, err := time.ParseDuration(string(*d))
+	if err != nil {
+		return nil, fmt.Errorf("invalid duration %q: %w", string(*d), err)
+	}
+	return &pd, nil
+}
+
+func validateNonNegative(name string, d *Duration) error {
```
jakobmoellerdev: this looks like its unnecessary, can there be a negative duration?

jakobmoellerdev: also the function can be inlined I believe, Validate is its only user afaik

api/utils/httpclient/transport.go
```diff
@@ -0,0 +1,71 @@
...
+	keepAlive, err := cfg.TCPKeepAlive.TimeDuration()
+	if err != nil {
+		return nil, err
+	}
+
+	// TCP Dialer settings
+	if dialTimeout != nil || keepAlive != nil {
+		// Clone() doesn't expose the original dialer, so we create a new one
+		// with the same defaults as http.DefaultTransport.
+		dialer := &net.Dialer{
```
jakobmoellerdev, line 34: shouldnt this be a global variable that is stored as package scope?

api/oci/config/httpconfig.go
piotrjanik: This is the standard lOCM config type pattern — `ApplyTo(_ cfgcpi.Context, target interface{})`.
I just followed it :)

  - https://github.com/open-component-model/ocm/blob/main/api/oci/config/type.go#L49                                                                                                                       
  - https://github.com/open-component-model/ocm/blob/main/api/credentials/config/type.go#L128                                                                                                              
  - https://github.com/open-component-model/ocm/blob/main/api/ocm/config/type.go#L75                                                                                                                       
  - https://github.com/open-component-model/ocm/blob/main/api/datacontext/config/logging/type.go#L70

api/oci/extensions/repositories/ocireg/repository.go
piotrjanik: Generally, yes, but it wasn't possible before, and I didn't want to add new features besides those in the task description.

api/oci/internal/httpsettings.go
piotrjanik: yes, negative values are allowed for tcp keep alive. `-1` disables it.

api/oci/extensions/repositories/ocireg/repository.go
piotrjanik: @Skarlso I added unit tests

api/oci/extensions/repositories/ocireg/repository.go
jakobmoellerdev: Arguably this is something that you should add since it even was flagged by the linter.

jakobmoellerdev (approved):

Skarlso (approved):

api/oci/extensions/repositories/ocireg/repository.go
piotrjanik: I think it was intentionally commented out because some registries at that time did not support TLS 1.3.

I just checked ECR, GHCR, GCR, Oracle and Quay, Docker and all support 1.3.

I will set it to 1.3 then

jakobmoellerdev:
please go through and resolve conversations

PR #1923, opened 2026-04-20
feat: close persistent references held on a context

Skarlso:

Closes dangling context references. All allocations are done all the time and finalizers don't truly close references. This means that everytime a context is created the ref is basically never cleaned up. This leads to the memory leak observed in the attached issue. This close will be called in the controller explicitly.

Related to https://github.com/open-component-model/ocm-project/issues/778.


api/credentials/internal/context.go
```diff
@@ -167,6 +167,10 @@ func newContext(configctx config.Context, reposcheme RepositoryTypeScheme, consu
 	}
 	c._InternalContext = datacontext.NewContextBase(c, CONTEXT_TYPE, key, configctx.GetAttributes(), delegates)
 	c.updater = cfgcpi.NewUpdaterForFactory(datacontext.PersistentContextRef(configctx), c.CredentialsContext)
+	c.Finalizer().With(func() error {
```
matthiasbruns, line 170: what happens if we call

```go
func (w *gcWrapper) SetContext(c *_context) {
	w._context = c
}
```

will the old context also finalize?

Skarlso: Yes.

jneisener (approved):

matthiasbruns (approved):

PR #1927, opened 2026-04-22
fix: use oci.ParseRef for ociArtifact input validation

fabianburth:
The ociArtifact input type validated its path using docker.ParseGenericRef, which splits on ":" and rejects references with more than two colon-separated parts. This broke OCI references containing digests (e.g. repo:tag@sha256:...) because the digest introduces a third colon.

Replace with oci.ParseRef, which is the same parser used at runtime by the ociArtifact access method.

### DIFF
--- a/cmds/ocm/commands/ocmcmds/common/inputs/types/ociartifact/spec.go
+++ b/cmds/ocm/commands/ocmcmds/common/inputs/types/ociartifact/spec.go
@@ -5,7 +5,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
-	"ocm.software/ocm/api/oci/extensions/repositories/docker"
+	"ocm.software/ocm/api/oci"
 	"ocm.software/ocm/api/oci/grammar"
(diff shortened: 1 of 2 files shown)


cmds/ocm/commands/ocmcmds/common/inputs/types/ociartifact/spec.go
```diff
@@ -39,8 +39,7 @@ func (s *Spec) Validate(fldPath *field.Path, ctx inputs.Context, inputFilePath s
 	allErrs = ValidateRepository(fldPath.Child("repository"), allErrs, s.Repository)
 	if s.Path != "" {
 		pathField := fldPath.Child("path")
-		_, _, err := docker.ParseGenericRef(s.Path)
-		if err != nil {
+		if _, err := oci.ParseRef(s.Path); err != nil {
```
matthiasbruns, line 42: can we add a test to prevent this from happening again?

matthiasbruns (approved):

Skarlso (approved):

PR #1946, opened 2026-05-06
fix: normalize Docker Hub single-component refs with library/ prefix

### LINKED ISSUE #1947: ParseRef missing library/ prefix for explicit docker.io single-component references
**Context**

`ParseRef` does not normalize `docker.io/bash:5.3.9` to `docker.io/library/bash:5.3.9`. The bare form `bash:5.3.9` is correctly normalized, but adding an explicit `docker.io` host bypasses the `library/` prefix logic.

**Version**

main (HEAD), also affects all released versions.

**To Reproduce**

```go
spec, _ := oci.ParseRef("docker.io/bash:5.3.9")
fmt.Println(spec.Repository) // prints "bash", expected "library/bash"
```

The input matches `ReferenceRegexp` (generic host/repo pattern) instead of `DockerLibraryReferenceRegexp` (which only handles bare names without a host).

**Actual behavior**

`docker.io/bash:5.3.9` → Host=`docker.io`, Repository=`bash`

**Expected behavior**

`docker.io/bash:5.3.9` → Host=`docker.io`, Repository=`library/bash`

**Additional Comments**

This causes failures in downstream tools that resolve the reference against Docker Hub without the `library/` prefix. Fix: #1946

knp-sap:
When docker.io is explicitly specified as the hostname with a single-component repository name (e.g. docker.io/bash:5.3.9), ParseRef now correctly normalizes it to docker.io/library/bash:5.3.9.

Previously, only bare image names (bash:5.3.9) received the library/ prefix via DockerLibraryReferenceRegexp. When docker.io was explicit, the reference matched ReferenceRegexp instead, which did not perform Docker Hub normalization.

Fixes: https://github.com/open-component-model/ocm/issues/1947


Skarlso:
Interesting. Did this cause any problems anywhere?

jakobmoellerdev (dismissed):

api/oci/ref.go
```diff
@@ -130,6 +130,9 @@ func ParseRef(ref string) (RefSpec, error) {
 		spec.Repository = string(match[3])
 		spec.Tag = pointer(match[4])
 		spec.Digest = dig(match[5])
+		if (spec.Host == dockerHubDomain || spec.Host == dockerHubLegacyDomain) && !strings.Contains(spec.Repository, "/") {
+			spec.Repository = "library" + grammar.RepositorySeparator + spec.Repository
+		}
```
matthiasbruns:
```suggestion
        needsDockerLibraryPath := pec.Host == dockerHubDomain || spec.Host == dockerHubLegacyDomain) && !strings.Contains(spec.Repository, "/")
        
        // inject library path segment into the repository part of the parsed ref.
        // for docker, images are resolved including a "library" path segment. This segment needs to be
        // added into the Repository field, otherwise the lookup would fail in case of docker's image registry
		if needsDockerLibraryPath {
			spec.Repository = "library" + grammar.RepositorySeparator + spec.Repository
		}
```

matthiasbruns: added docs & a more understandable naming

matthiasbruns (approved):

frewilhelm (approved):
In theory, one should rather use a `regex` to sort that out, e.g. like its done with `DockerLibraryReferenceRegexp` or `DockerReferenceRegexp`. But these are pretty much black magic and this fix works as well and won't produce any unexpected behaviour.

PR #1976, opened 2026-05-29
chore: use github environment

frewilhelm:

Use GitHub environment in workflows

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,6 +27,9 @@ jobs:
   check:
     name: Check Release Preconditions
     runs-on: ubuntu-latest
+    environment:
+      name: release
+      deployment: false
     permissions:
       contents: read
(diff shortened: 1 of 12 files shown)


.github/workflows/release.yaml
```diff
@@ -103,6 +106,9 @@ jobs:
     - components
     name: Release Build
     runs-on: large_runner
+    environment:
+      name: release
```
matthiasbruns, line 110: so we also want to call this automation to keep it in sync with the mono-repo?

frewilhelm: No, the environments are not related to each other. I called it `release` here because its related to the releases. In the mono-repo we need two different environments because one is gated and the other is not

matthiasbruns (approved):

Skarlso (approved):

PR #1996, opened 2026-06-08
chore: add env-var gated repository-exist call

frewilhelm:

We saw one failure of oci artifact transfers that we were not able to reproduce. We only know that reverting this change that was introduced by #1676 is fixing the issue. We do not want to revert it exactly because other users are using the concurrent workerpool and it just works fine. Hence, we introduced an environment variable gate, so users need to actively opt-in.


jakobmoellerdev (dismissed):

api/tech/oras/pusher.go
```diff
@@ -48,19 +64,46 @@ func (c *OrasPusher) Push(ctx context.Context, d ociv1.Descriptor, src Source) (
...
 	}
 
 	return nil
 }
+
+// pushExistsCheckEnabled reports whether the optional pre-push Exists() check
+// is requested via PushExistsCheckEnvVar. Unset, empty, or unparsable values
+// disable the check (default behaviour).
+func pushExistsCheckEnabled() bool {
+	v, ok := os.LookupEnv(PushExistsCheckEnvVar)
```
matthiasbruns: the call `LookupEnv` is not really needed

`os.GetEnv` returns an empty `string`, which will always be parsed to `false` with `strconv.ParseBool(v)`

jakobmoellerdev: if we want to be nit-gritty like this, we should change it so that the env here is looked up exactly once per app start because there is no need to react to a changing env var

matthiasbruns: actually... :D 
it's somewhat cached

`syscall/env_unix.go`

```go
var copyenv = sync.OnceFunc(func() {
	env = make(map[string]int)
	for i, s := range envs {
		for j := 0; j < len(s); j++ {
			if s[j] == '=' {
				key := s[:j]
				if _, ok := env[key]; !ok {
					env[key] = i // first mention of key
				} else {
					// Clear duplicate keys. This permits Unsetenv to
					// safely delete only the first item without
					// worrying about unshadowing a later one,
					// which might be a security problem.
					envs[i] = ""
				}
				break
			}
		}
	}
})

func Getenv(key string) (value string, found bool) {
	copyenv()
	if len(key) == 0 {
		return "", false
	}

	envLock.RLock()
	defer envLock.RUnlock()

	i, ok := env[key]
	if !ok {
		return "", false
	}
	s := envs[i]
	for i := 0; i < len(s); i++ {
		if s[i] == '=' {
			return s[i+1:], true
		}
	}
	return "", false
}
```

matthiasbruns (dismissed):
no need to be too nit picky on this one

frewilhelm:
> no need to be too nit picky on this one

no, it was a fair point from both of you. I don't mind!

api/tech/oras/pusher.go
```diff
@@ -2,16 +2,46 @@ package oras
...
+
+// pushExistsCheckEnabled reports whether the optional pre-push Exists() check
+// is requested via PushExistsCheckEnvVar. The env var is consulted exactly
+// once at package initialisation; unset, empty, or unparsable values disable
+// the check.
+//
+// strconv.ParseBool returns (false, err) for "" - the os.Getenv result for an
+// unset variable — so a single call covers the unset and unparsable cases.
+var pushExistsCheckEnabled = func() bool {
+	enabled, _ := strconv.ParseBool(os.Getenv(PushExistsCheckEnvVar))
```
jakobmoellerdev: this should be a sync.OnceValues

jakobmoellerdev (approved):

PR #2009, opened 2026-06-18
chore(releases-045): cherry-pick: add env-var gated repository-exist call (#1996)

frewilhelm:
Cherry-pick of #1996  from main to releases/0.45, so the fix is available for testing in the release candidate.


frewilhelm:
Opened against the wrong branch. Closing and reopening it

PR #2032, opened 2026-07-07
chore(deps): bump fulcio; downgrade crypto11

frewilhelm:
- bump fulcio only possible when removing `replace` directive which downgrades `crypto11` for module `cosign`
  - Fulcio changed the import path for `crypto11` to the new one while `cosign` still imports the old one. Because of the `replace` directive, this breaks the module bump to `1.6.2` because the dependency depends on two paths
- [Diff](https://github.com/thales-transfer/crypto11/compare/v1.2.5...v1.6.2) between `crypto11` `1.2.5` and `1.6.2`
  - No important changes or vulnerabilities affect us, check e.g. `govulncheck -mode=source ./...`


frewilhelm:
Superseded by #2040

PR #2112, opened 2026-09-14
chore(deps): bump github.com/gobwas/glob from 0.2.3 to 1.0.0 (fix 2109)

frewilhelm:
Fixes #2109

The new release introduces a breaking change replacing the `Glob` type with `Pattern`.

We might postpone this upgrade as there is no security vulnerability or the like that requires us to upgrade.


frewilhelm:
`golangci-lint` has not catched up yet. Until then, we cannot merge this PR

